PR #367, opened 2026-04-25
feat(preempt): match skill catalog and fix background/mcp discovery

lmorchard:
## Summary

- Mark `background` and `mcp` skills as `always-loaded: true` so their tools are active from turn 1, matching what CLAUDE.md already (incorrectly) claimed about them.
- Extend the pre-emptive search to also match the discovered skill catalog and surface likely-relevant non-activated skills as a short `<preempt_skill_hint>` system message — hint-only, no auto-activation.

## Why

Triggered by a painful exchange where the agent tried to start a blog generator in the background and went through: wrong workspace_search params (twice) → call shell_background_start → "unknown tool" → tool_search (no result) → activate_skill background → finally start it. The postmortem identified the symptoms but missed two real causes:

1. CLAUDE.md claimed both `background` and `mcp` skills "auto-activate", but their SKILL.md files lacked `always-loaded: true` (only `vault` had it). They had to be activated on demand.
2. The pre-emptive tool search candidate pool only contained tools from already-activated skills, so `shell_background_start` was invisible to keyword matching even though "background" was in the user message. The mechanism designed to catch exactly this case couldn't see it.

This PR fixes both. Always-loaded handles those two specific skills directly; the new skill-catalog match handles any other on-demand skill the same way going forward.

## What changes

**Skill frontmatter:**
- `src/decafclaw/skills/background/SKILL.md` — add `always-loaded: true`
- `src/decafclaw/skills/mcp/SKILL.md` — add `always-loaded: true`

**Pre-emptive skill match:**
- `src/decafclaw/context.py` — `SkillState.preempt_matches: set[str]`
- `src/decafclaw/context_composer.py` — `_compose_preempt_skill_matches()` runs alongside the existing tool match. Skills already in `ctx.skills.activated` (including always-loaded) are excluded from the candidate pool. Matches render as a small system message:
  ```
  <preempt_skill_hint>
  These skills look relevant to the current message: project, ingest.
  Their tools are NOT loaded yet — call activate_skill(name) before trying to use any of their tools.
  </preempt_skill_hint>
  ```
  Hint-only — no auto-activation, so the existing user-confirmation flow for non-`auto-approve` skills is untouched.

**Docs:**
- `CLAUDE.md` — switch background/mcp key-files entries from "auto-activates" to canonical "always-loaded"; expand the pre-emptive search summary to describe the skill match.
- `docs/preemptive-tool-search.md` — new "Skill catalog match" section, plus an intro mention.

**Tests:**
- `tests/test_context_composer.py::TestComposePreemptSkillMatches` — 8 tests covering disable, empty input, basic match + hint shape, activated-skill exclusion, empty catalog, max_matches cap, fresh-per-turn reset, prior-assistant continuity.

## Test plan

- [x] `make lint` clean
- [x] `make typecheck` clean (0 errors, 0 warnings)
- [x] `make test` — 1902 passed in 10.9s (8 new)
- [ ] Live test in the web UI: ask the bot to "run my blog generator in the background" — verify it can call `shell_background_start` directly without an `activate_skill` round-trip.
- [ ] Live test: ask about a non-always-loaded skill (e.g. `!ingest` topic) — verify the `<preempt_skill_hint>` shows up in the context inspector / sidecar diagnostics.


lmorchard:
Addressed both Copilot review comments in ee86542:

1. Added an early-return guard for `max_matches <= 0` so a misconfig can't produce a hint with an empty name list and a bogus SourceEntry. Mirrors `preempt_search.match_tools`.
2. Run `html.escape` on each skill name before joining into the `<preempt_skill_hint>` wrapper so a workspace skill named `</preempt_skill_hint>...` (or with a stray newline) can't break out and inject prompt content.

Added regression tests for both. Lint and full suite green (1904 passed).

PR #404, opened 2026-04-27
refactor(conversation_manager): PersistedTurnState single source of truth

### LINKED ISSUE #378: conversation_manager: parallel save/restore for persisted turn state needs single source of truth
## Context

Surfaced during the audit alongside [#377](https://github.com/lmorchard/decafclaw/issues/377) (vault provenance) — really, alongside the `Context.fork_for_tool_call` fix on branch `fix/scheduled-task-mode-fork-propagation`. That branch fixed three "laundry list" copy patterns (`fork_for_tool_call`, `NotificationRecord` round-trip, `ConfirmationRequest`/`Response` round-trip) and added a CLAUDE.md convention banning the pattern.

This issue covers the one audit finding deliberately **left out** of that branch because it's a different bug shape and warrants its own design pass.

## The pattern

`ConversationManager._save_conversation_state` and `_restore_per_conv_state` are a parallel save/restore pair (`src/decafclaw/conversation_manager.py:735-756`). Save copies a few `ctx` fields onto a `state.skill_state: dict | None` plus directly onto `state.skip_vault_retrieval`; restore reads them back. They look like this:

```python
def _save_conversation_state(self, state, ctx):
    if ctx.skills.activated:
        state.skill_state = {
            \"extra_tools\": ctx.tools.extra,
            \"extra_tool_definitions\": ctx.tools.extra_definitions,
            \"activated_skills\": ctx.skills.activated,
        }
    if ctx.skip_vault_retrieval:
        state.skip_vault_retrieval = True

def _restore_per_conv_state(self, state, ctx):
    if state.skill_state:
        ctx.tools.extra = state.skill_state.get(\"extra_tools\", {})
        ctx.tools.extra_definitions = state.skill_state.get(
            \"extra_tool_definitions\", [])
        ctx.skills.activated = state.skill_state.get(
            \"activated_skills\", set())
    if state.skip_vault_retrieval:
        ctx.skip_vault_retrieval = True
    if state.active_model:
        ctx.active_model = state.active_model
```

## Why it's a different bug class

The three patterns we already fixed all had the same shape: copy the source object's fields onto a target with no further filtering. `vars()` / `dataclasses.fields()` iteration is the right idiom because we want **every** field to propagate.

The save/restore pair is structurally different. Only a *deliberate subset* of `Context` persists across turns — most ctx state (per-call counters, per-call tool IDs, transient flags) explicitly should NOT survive. So `vars()` iteration would do the wrong thing. The risk is the **two parallel field lists** drifting out of sync — adding a persistent field requires both a save edit and a restore edit, and the symmetry is enforced only by code review.

There's already at least one inconsistency: `_restore_per_conv_state` reads `state.active_model` and applies it to ctx, but `_save_conversation_state` doesn't write `state.active_model` (it's set by the web \`set_model\` handler, not by turn state — but the asymmetry between save and restore is exactly the kind of thing that should be reflected in the structure).

## Proposed direction

Introduce a typed dataclass that represents \"what persists\" and becomes the single source of truth:

```python
@dataclass
class PersistedTurnState:
    extra_tools: dict[str, Any] = field(default_factory=dict)
    extra_tool_definitions: list[dict] = field(default_factory=list)
    activated_skills: set[str] = field(default_factory=set)
    skip_vault_retrieval: bool = False
    active_model: str = \"\"
```

Then save and restore each map ctx ↔ this dataclass through symmetric helpers. Adding a new persistent field means: add it to `PersistedTurnState`, wire one save line, wire one restore line — and a test that walks `dataclasses.fields(PersistedTurnState)` flags either side dropping the field.

Replaces `state.skill_state: dict | None` with `state.persisted: PersistedTurnState`. Self-contained refactor — the field is only read/written inside `conversation_manager.py`.

## Scope

- `_save_conversation_state` and `_restore_per_conv_state` rewritten to map through `PersistedTurnState`.
- `ConversationState.skill_state` dict replaced (or kept as a `@property` for one release if anything outside the module reads it; current grep shows nothing does).
- Existing test `test_enqueue_turn_wake_kind_restores_skill_state` updated to construct a `PersistedTurnState` instead of poking the dict shape.
- New test that walks `dataclasses.fields(PersistedTurnState)` and verifies both save and restore touch every field — turns the audit's structural concern into a regression test.

Audit-confirmed scope: `state.skill_state` is only referenced inside `conversation_manager.py` (5 lines, 2 methods), so no cross-module ripples.

## Project board

- **Priority:** P0
- **Size:** S (focused refactor + 1 test rewrite + 1 new test)

lmorchard:
## Summary

- Adds `PersistedTurnState` dataclass + `_PERSISTED_BINDINGS` table as the single source of truth for what persists across agent turns.
- Replaces the previous trio (`skill_state: dict | None`, `skip_vault_retrieval: bool`, `active_model: str`) on `ConversationState` with one `persisted: PersistedTurnState` field.
- `_save_conversation_state` and `_restore_per_conv_state` walk the same binding table; adding a new persisted field can no longer silently drop out of either side.

## Why

`#378` flagged this pair as a different-shape bug from the laundry-list copy patterns we already fixed. CLAUDE.md says "never enumerate fields when copying" but that idiom assumes you want every field — here we want a deliberate subset, so `vars()` iteration would be wrong. The fix is a typed dataclass naming what persists plus a binding table walked symmetrically by both halves.

## Design notes

- `_CTX_DRIVEN_FIELDS` distinguishes fields whose value flows ctx → state on save (extra_tools / extra_tool_definitions / activated_skills / skip_vault_retrieval) from externally-driven ones (`active_model`, set via `set_flag` from web UI). Save never overwrites externally-driven fields.
- Truthy-only writes preserve the sticky-once-set semantics the inline conditionals had: a False ctx value never clobbers a True persisted value.
- `set_flag` write-through is automatic for any field that's a `PersistedTurnState` field name — keeps `web/websocket.py` callers (`set_flag("active_model", ...)`) working unchanged.

## Test plan

- [x] `test_persisted_field_bindings_exhaustive` — fields ↔ bindings symmetric
- [x] `test_ctx_driven_fields_subset_of_persisted` — typo guard on `_CTX_DRIVEN_FIELDS`
- [x] `test_save_restore_round_trip` — populate every field, save, restore on fresh ctx, all values flow through
- [x] `test_save_does_not_overwrite_externally_driven_fields` — `active_model` set via `set_flag` survives save with a different ctx value
- [x] `test_set_flag_writes_through_to_persisted` — write-through correctness
- [x] `test_save_truthy_only_preserves_sticky_semantics` — sticky-once-set guard
- [x] Existing `test_enqueue_turn_wake_kind_restores_skill_state` updated to construct `PersistedTurnState` directly — passes without behavior change
- [x] Full pytest run (2205 tests) clean

Closes #378

### DIFF
--- a/docs/dev-sessions/2026-04-27-1654-persisted-turn-state/spec.md
+++ b/docs/dev-sessions/2026-04-27-1654-persisted-turn-state/spec.md
@@ -1,0 +1,157 @@
+# PersistedTurnState — single source of truth for conversation persistence (#378)
+
+## Problem
+
+`ConversationManager._save_conversation_state` and `_restore_per_conv_state` are a save/restore pair (`conversation_manager.py:765-790`). They copy a deliberate subset of `Context` state onto `ConversationState` and back. The bug shape: **two parallel hand-maintained field lists** that drift out of sync silently — adding a persistent field requires both a save edit and a restore edit, with symmetry enforced only by code review.
+
+CLAUDE.md already bans the "hand-list fields when copying" pattern, but `_save`/`_restore` carved out an exception because not every ctx field should persist (per-call counters, per-call IDs, transient flags must NOT survive). The right fix isn't `vars()` iteration; it's a typed dataclass that names "what persists" plus a single binding table that walks fields symmetrically.
+
+## Why this isn't the same fix as #377
+
+The audit fixes that #377 addressed (`fork_for_tool_call`, notification round-trip, confirmation round-trip) all wanted *every* field to propagate — `vars()` / `dataclasses.fields()` iteration was the right idiom there.
+
+This pair wants only a *deliberate subset*. So `vars()` would be wrong; we need a structurally-enforced subset.
+
+## Design
+
+### `PersistedTurnState` dataclass
+
+```python
+@dataclass
+class PersistedTurnState:
+    """Per-conversation state that persists across turns.
+
+    Two field categories live here:
+    - Ctx-driven: written by `_save_conversation_state` from ctx,
+      read by `_restore_per_conv_state` onto ctx.
+    - Externally-driven: set by `manager.set_flag()` from web /
+      transport handlers, read by restore onto ctx; save never
+      overwrites these.
+    """
+    extra_tools: dict = field(default_factory=dict)
+    extra_tool_definitions: list = field(default_factory=list)
+    activated_skills: set = field(default_factory=set)
+    skip_vault_retrieval: bool = False
+    active_model: str = ""
+```
+
+Replaces the existing trio on `ConversationState`:
+- `skill_state: dict | None` (the dict shape that today bundles `extra_tools` / `extra_tool_definitions` / `activated_skills`)
+- `skip_vault_retrieval: bool`
+- `active_model: str`
+
+…with a single field `persisted: PersistedTurnState = field(default_factory=PersistedTurnState)`.
+
+### Binding table — single source of truth
+
+```python
+# Per-field reader/writer pair, exhaustive over PersistedTurnState
+# fields. Adding a field to PersistedTurnState requires adding an
+# entry here (the regression test in test_conversation_manager.py
+# enforces this).
+_PERSISTED_BINDINGS: dict[str, tuple[Callable, Callable]] = {
+    "extra_tools": (
+        lambda ctx: ctx.tools.extra,
+        lambda ctx, v: setattr(ctx.tools, "extra", v),
+    ),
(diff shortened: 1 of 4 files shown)


lmorchard:
@copilot — addressed your inline note in 0aeadab: added module-level `_PERSISTED_FIELD_NAMES` frozenset (precomputed from `dc_fields(PersistedTurnState)` at import) and `set_flag` now uses it instead of recomputing the set on every call. Stays in sync with the dataclass automatically since it's derived.

Issue #358, opened 2026-04-24
Agent-authored web UI content: workspace widgets + iframe sandbox


lmorchard:
Follow-up to #256 (widget catalog & canvas panel).

## Context

#256 ships a widget catalog with two trust tiers:

1. `src/decafclaw/web/static/widgets/` — bundled, developer-authored
2. `data/{agent_id}/widgets/` — admin-authored (outside agent write path)

It deliberately **drops the originally-spec'd third tier** (`workspace/widgets/`) for v1 because `workspace/` is agent-writable, and `widget.js` is code that runs in an authenticated browser context. Giving the agent the ability to drop JS into a workspace subdirectory is a privilege-escalation surface (session-token exfil, arbitrary `/api/workspace/*` reads, DOM spoofing, etc.).

This issue tracks the two related but deferred capabilities.

### A. Workspace-tier widgets

Let `workspace/widgets/` be a first-class catalog location, with a trust model that makes agent-authored widgets safe to load.

Possible approaches (all deferred for evaluation):

- **First-use approval pattern**, mirroring skill permissions. The first time a new `workspace/widgets/foo/` appears, the user approves (or rejects) loading its `widget.js`. Approvals stored in `data/{agent_id}/widget_permissions.json`.
- **Split trust**: `widget.json` descriptors can live in workspace, but `widget.js` must live in `data/{agent_id}/widgets/`. Descriptor can be agent-authored, renderer cannot.
- **Signed widgets**: widget.js hash must be approved by admin before loading.

### B. Agent-authored iframe sandbox widget

The original #256 spec's \"Future extensions\" section calls for an `iframe_sandbox` widget type whose Lit component renders an iframe with agent-generated HTML/JS/CSS:

- CSP-locked, no network access by default
- postMessage communication with the host page
- Same WidgetRequest protocol, different renderer
- Safe-by-construction even when content is arbitrary agent output

This is probably the safer path to \"agent-authored rich UI\" than loading agent-authored Lit components directly.

## Out of scope here

- Widget hot-reload (covered in #256 future extensions, separate concern)
- Web Applets protocol interop (also covered in #256 future extensions)

## Prerequisites

#256 (widget catalog) must ship first — this issue builds on the two-tier catalog it establishes.

PR #423, opened 2026-04-29
feat(widgets): iframe_sandbox — agent-authored content (#358)

lmorchard:
Closes part B of #358. Part A (workspace-tier widgets with first-use approval) stays deferred.

## Summary

- Bundled `iframe_sandbox` widget renders agent-authored HTML/CSS/JS in a CSP-locked, sandboxed iframe with no network access.
- Two layers of defense: `sandbox=\"allow-scripts\"` (NOT `allow-same-origin` — null origin denies parent storage/cookies) + backend-injected CSP meta (`default-src 'none'`, blocking all network).
- Agent supplies `{body, title?}`; backend `_normalize_iframe_sandbox` wraps it into a full document with the CSP meta. Frontend just sets `iframe.srcdoc`.
- No new tool — agent emits via existing `canvas_new_tab(widget_type=\"iframe_sandbox\", data={...})` or any tool returning a `WidgetRequest`.
- Adds a per-widget `normalize(name, data)` hook on `WidgetRegistry` (module-level `_NORMALIZERS` dispatch). Pure, idempotent functions. Currently only `iframe_sandbox` registers one; easy to extend.

## Why backend normalization (vs frontend wrap)

Discussed up front: even a buggy or future malicious admin/workspace-tier `widget.js` cannot drop the CSP, because by the time wrapped HTML reaches the browser the meta is already in the document. Single canonical wrapper, no divergence.

## Idempotence

Re-normalizing the same data produces identical output, and a stale `html` field in input is overwritten by re-wrapping from `body`. Verified by `test_iframe_sandbox_normalize_idempotent`. Matters because `canvas_read` returns the wrapped form; if the agent round-trips that into `canvas_update`, normalization regenerates `html` from `body` rather than replaying the (potentially poisoned) prior `html`.

## Caps

256 KB hard cap on `body` via JSON Schema `maxLength`. Title 200 chars, HTML-escaped before injection (`</title>` neutralized).

## Out of scope (filed as future)

- Workspace-tier widgets with first-use approval — #358 part A still open.
- postMessage→widget-response from inside iframe_sandbox — future. Trust check at that point: `event.source === iframe.contentWindow` (origin is `null` for sandboxed-without-same-origin iframes, so we can't use origin).

## Test plan

- [x] Unit: registry parses descriptor; modes inline+canvas; accepts_input false.
- [x] Unit: schema rejects missing/oversized body; permits round-trip with `html` field.
- [x] Unit: normalize injects CSP meta, escapes title, preserves body+title alongside html.
- [x] Unit: normalize is idempotent; stale html overwritten.
- [x] Unit: canvas `new_tab` and `update_tab` invoke normalize after validate (verified with a fake normalizer).
- [x] `make check` clean (ruff, pyright, tsc).
- [x] `make test` clean (2240 passed, ~19s).
- [ ] **Browser smoke** — pending. Drop a small interactive demo via `canvas_new_tab(widget_type=\"iframe_sandbox\", data={\"body\": \"<h1>hi</h1><script>document.body.style.background='lime'</script>\", \"title\": \"demo\"})` and verify: (1) renders, (2) script runs, (3) `fetch('https://example.com')` blocked by CSP, (4) `parent.location` throws (sandbox null-origin).

### DIFF
--- a/docs/dev-sessions/2026-04-28-2236-iframe-sandbox-widget/notes.md
+++ b/docs/dev-sessions/2026-04-28-2236-iframe-sandbox-widget/notes.md
@@ -1,0 +1,38 @@
+# iframe_sandbox session — notes
+
+## What shipped
+
+Closes #358 part B. Part A (workspace-tier widgets with first-use approval) stays deferred.
+
+A bundled `iframe_sandbox` widget that renders agent-authored HTML/CSS/JS in a CSP-locked, sandboxed iframe with no network access. Same WidgetRequest plumbing as existing widgets — no new tool needed; the agent uses `canvas_new_tab(widget_type="iframe_sandbox", data={"body": "...", "title": "..."})`.
+
+## Design decisions worth remembering
+
+- **Two layers of defense.** `sandbox="allow-scripts"` (NOT `allow-same-origin`) on the iframe + `<meta http-equiv="Content-Security-Policy">` injected backend-side. Either alone wouldn't be enough: sandbox without CSP allows network, CSP without sandbox allows storage/cookie access from inside the frame.
+- **CSP injection lives in Python.** `_normalize_iframe_sandbox` in `widgets.py` produces the wrapped doc. The backend sends the wrapped HTML over WS; widget.js just sets `iframe.srcdoc`. This means even a buggy or malicious admin/workspace-tier `widget.js` can't drop the CSP — by the time wrapped HTML reaches the browser the meta is already in the document.
+- **Per-widget normalization hook on the registry.** Added `WidgetRegistry.normalize(name, data)` + a module-level `_NORMALIZERS` dispatch dict. Pure functions, idempotent. Currently only `iframe_sandbox` registers one. Easy to extend later without touching the class.
+- **Idempotent normalization.** A stale `html` field in input is overwritten by re-wrapping from `body`, so a `canvas_read` → `canvas_update` round-trip can't compound or replay a poisoned document. Verified by `test_iframe_sandbox_normalize_idempotent`.
+- **Data shape diverges between input and stored.** Input: `{body, title?}`. Stored/transmitted: `{body, title?, html}`. The schema lists `html` in `properties` (so round-trip doesn't fail validation under `additionalProperties: false`) but doesn't require it. The schema validates input shape; widget.js consumes the post-normalization shape.
+- **256 KB body cap.** Enforced via `maxLength` in the data_schema. Big enough for non-trivial demos (small p5.js sketches, charts) without unbounded archive growth.
+- **No widget-response in v1.** `accepts_input: false`. To add later: validate `event.source === iframe.contentWindow` in a postMessage handler — origin will be `null` for sandboxed-without-same-origin so source-equality is the trust check.
+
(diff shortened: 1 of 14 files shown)


lmorchard:
## Browser smoke test — passed

Exercised the live widget via Playwright MCP against `make dev` (port 18880). Pushed an `iframe_sandbox` widget into the canvas via `POST /api/canvas/{conv}/new_tab`. The widget body included three inline-script probes that report status to its own DOM.

**Three sandbox/CSP behaviors all confirmed (read from inside the iframe):**

- ✅ **Section A — script ran.** Inline `<script>` executed; the body's first DOM update reported `OK — script executed`.
- ✅ **Section B — fetch blocked.** `fetch('https://example.com/')` rejected with `Failed to fetch`. Browser console additionally logged: *"Connecting to 'https://example.com/' violates the following Content Security Policy directive: `default-src 'none'`. Note that 'connect-src' was not explicitly set, so 'default-src' is used as a fallback. The action has been blocked."*
- ✅ **Section C — parent.location blocked.** Reading `parent.location.href` from inside the iframe threw: *"Blocked a frame with origin `null` from accessing a cross-origin frame."* — confirms the sandbox null-origin denial.

**Parent → iframe access (read by Playwright on the host page):**

- `iframe.sandbox.value === "allow-scripts"` (no `allow-same-origin`)
- `iframe.referrerPolicy === "no-referrer"`
- `iframe.srcdoc` starts with `<!doctype html>`, contains the CSP `<meta>` with `default-src 'none'`
- `iframe.contentDocument === null` — parent can't read iframe DOM
- `iframe.contentWindow.location.href` throws — parent can't read iframe location either
- `iframe.contentWindow` itself is reachable (so `postMessage` would work for future input-widget mode)

Screenshot saved at `docs/dev-sessions/2026-04-28-2236-iframe-sandbox-widget/smoke-test.png` (not committed — it's a local artifact for the dev session).

Round-trip safety (`canvas_update` regenerates `html` from fresh `body`, ignoring stale `html`) is covered by the unit test `test_update_tab_runs_normalize` and is agent-tool-only — no REST endpoint to drive it from the browser.

lmorchard:
Addressed all four Copilot comments in 25ad5d7:

1. **`html` maxLength** (widget.json) — Added `maxLength: 524288` (512 KB, 2x body cap) to the `html` field. Closes the loophole where a tiny `body` + huge `html` would pass validation before normalize wipes it.
2. **Weak `https:`/`https://` assertion** (test_widgets.py) — Replaced with the CSP-tag-scoped check Copilot suggested: extract the CSP meta substring and assert AND across both schemes plus `fetch-src`. Body text can no longer accidentally satisfy it.
3. **Spec drift** (spec.md) — Updated to match shipped schema: `additionalProperties: false`, `html` listed in `properties` to allow round-trips.
4. **Admin-tier normalizer leak** (widgets.py) — Gated `WidgetRegistry.normalize` on `descriptor.tier == "bundled"`. An admin widget that shares a bundled name no longer inherits the bundled normalizer (its data shape may be entirely different). New test `test_iframe_sandbox_normalizer_skips_admin_override` covers the override path.

`make check` clean, `make test` passes (2241 tests).

PR #424, opened 2026-04-29
fix(llm/vertex): strip JSON-Schema keys Vertex rejects in tool params

lmorchard:
## Summary

- Vertex tool-parameter schema is a restricted subset of OpenAPI 3. MCP-provided tool schemas (Playwright + likely others) sometimes use full JSON-Schema keywords like \`propertyNames\`, \`patternProperties\`, \`\$defs\`, \`dependentRequired\`, etc. — Vertex rejects them with HTTP 400 \`Unknown name \"X\"\`, killing the agent turn mid-execution.
- \`_clean_schema\` previously stripped only \`\$schema\`. Extend to a denylist of all the JSON-Schema keywords Vertex doesn't accept and recurse through \`oneOf\`/\`anyOf\`/\`allOf\` so combinator branches get scrubbed too.
- Denylist not allowlist, so legitimate constraints (\`minLength\`, \`enum\`, \`maxItems\`, …) keep flowing through. Add keys to \`_VERTEX_UNSUPPORTED_KEYS\` as new rejection cases surface.

## Stripped keywords

- Schema metadata: \`\$schema\`, \`\$id\`, \`\$defs\`, \`definitions\`, \`\$ref\`
- Property-shape: \`propertyNames\`, \`patternProperties\`, \`dependencies\`, \`dependentRequired\`, \`dependentSchemas\`
- Conditional branches: \`if\`, \`then\`, \`else\`
- Numeric: \`multipleOf\`

## Repro that prompted this

A user turn dispatched to Vertex died with:

> Invalid JSON payload received. Unknown name \"propertyNames\" at 'tools[0].function_declarations[41].parameters.properties[3].value': Cannot find field.

Index 41 was an MCP tool whose input schema used \`propertyNames\`. After the fix, the schema is sanitized before serialization and the request goes through.

## Test plan

- [x] New unit test (\`test_tool_schema_strips_propertynames_and_pattern_keywords\`) builds a request body with an adversarial schema covering every newly-stripped keyword and asserts they're absent post-translation while legitimate keys (\`type\`, \`required\`) survive.
- [x] \`make check\` clean (ruff + pyright + tsc).
- [x] \`make test\` clean (2233 passed).
- [ ] Live verification: re-run the failed turn shape and watch for a clean Vertex 200.

### DIFF
--- a/src/decafclaw/llm/providers/vertex.py
+++ b/src/decafclaw/llm/providers/vertex.py
@@ -340,9 +340,35 @@ class _VertexStreamState:
 # Message format translation
 # ---------------------------------------------------------------------------
 
+_VERTEX_UNSUPPORTED_KEYS = frozenset({
+    # Schema metadata Vertex doesn't process.
+    "$schema", "$id", "$defs", "definitions", "$ref",
+    # JSON Schema keywords for property-name / pattern-shape constraints —
+    # not in the OpenAPI 3 subset Vertex accepts.
+    "propertyNames", "patternProperties", "dependencies", "dependentRequired",
+    "dependentSchemas",
+    # Conditional schema branches.
+    "if", "then", "else",
+    # Numeric constraint Vertex rejects on some schema versions.
+    "multipleOf",
+})
+
+
 def _clean_schema(schema: dict) -> dict:
-    """Remove keys that Vertex/Gemini doesn't accept in JSON Schema (e.g. $schema)."""
-    cleaned = {k: v for k, v in schema.items() if k != "$schema"}
+    """Strip JSON-Schema keywords Vertex/Gemini doesn't accept.
+
+    Vertex's tool-parameter schema is a restricted subset of OpenAPI 3.
+    MCP-provided tool schemas (e.g. Playwright) frequently use full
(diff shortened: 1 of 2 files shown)


lmorchard:
Addressed both Copilot comments in 660407b:

1. **Coverage gap** — extended the adversarial schema to include every current member of \`_VERTEX_UNSUPPORTED_KEYS\` (\`\$id\`, \`definitions\`, \`\$ref\`, \`dependentSchemas\` were missing) with explicit assertions for each. Added a defense-in-depth check at the end: the test asserts every key in the denylist appears somewhere in the test schema, so adding a key to \`_VERTEX_UNSUPPORTED_KEYS\` without extending coverage fails this test loudly instead of silently.
2. **Combinator recursion untested** — new test \`test_tool_schema_scrubs_unsupported_keys_inside_combinator_branches\` buries \`propertyNames\` / \`patternProperties\` / \`\$defs\` inside nested \`oneOf\` / \`anyOf\` / \`allOf\` branches and asserts each gets scrubbed.

Same commit cherry-picked into #423 since that branch has a duplicate of this fix for local unblock.

PR #434, opened 2026-04-30
Fix #393: separate history from current-turn injections in compose()

### LINKED ISSUE #393: Token accounting underestimates archived role-remap messages
Surfaced during PR #392 review (Copilot). Pre-existing issue, slightly extended by adding ``conversation_notes`` to the role-remap dict in ``ContextComposer``.

## Current behavior

\`ContextComposer.compose()\` builds an LLM-facing message list by remapping a set of internal roles (\`vault_retrieval\`, \`vault_references\`, and as of #392 also \`conversation_notes\`) to \`user\` so providers accept them.

The fixed-cost / dynamic-budget calculation:
1. Reserves token budget for the messages **injected this turn** (each tracked via a \`SourceEntry\`).
2. Sums tokens for "existing history" by iterating \`history\` and summing content lengths.

The bug: prior turns' archived messages with remap roles (\`vault_retrieval\`, \`vault_references\`, \`conversation_notes\`) **are** in \`history\` and **do** get sent to the LLM (after remap), but they're filtered out of the \`history_only\` token estimate because that estimate assumes their tokens were already accounted for in a current-turn \`SourceEntry\`. They were — but only for the turn they were injected. On subsequent turns nothing accounts for them.

## Impact

Multi-turn conversations underestimate prompt tokens by roughly the cumulative sum of all auto-injected wiki / memory / notes content from prior turns. The dynamic memory budget over-allocates, and ``show_context_status`` / sidecar diagnostics under-report.

In practice the per-turn drift is small (a few hundred tokens for vault refs, ≤4 KB for notes), but it accumulates linearly across turns until compaction.

## Possible fixes

- **Track injected-this-turn message ids and exclude only those** from the history estimate. The current code already builds \`injected_ids\` — extend the same idea consistently across all remap roles, then no separate role-based filter.
- **Include remap-role messages in \`history_entry.tokens_estimated\`** but flag them so they're not double-counted with the current-turn SourceEntry.

The first is simpler and matches what the comment in #392 suggests.

## Why this isn't part of #392

Pre-existing issue (\`vault_retrieval\` / \`vault_references\` already had it). Adding \`conversation_notes\` to the remap dict makes the same bug apply to one more role; fixing it cleanly means revisiting the broader \`history_entry\` calculation, which is a bigger change than #392 should carry.

## Related

- #392 — added \`conversation_notes\` to \`role_remap\`
- \`src/decafclaw/context_composer.py::compose\` — \`history_only\` filter and \`history_entry\` calculation

lmorchard:
## Summary

- Fixes #393: archived `vault_retrieval` / `vault_references` / `conversation_notes` messages are now correctly counted in `history_entry.tokens_estimated`, the sidecar `total_tokens_estimated`, and the `[Context: ...]` status line.
- Restructures `ContextComposer.compose()` so the input `history` list is read-only during all token-budget and diagnostics calculation. Eliminates the dual filter (id-based + role-based) that produced the bug — the role-based filter quietly excluded archived auto-injected role messages even though the LLM still sees them after remap.

Real-data smoke against an archived 46-message conversation: pre-fix reported 2623 tokens for the history source entry; post-fix reports 9569 — a 6946-token (~73%) correction.

## Design Decisions

- **`history` is read-only during all calculation.** The four mid-compose `history.append` calls are replaced by a single block at the end (after diagnostics), preserving the load-bearing side effect that the agent loop relies on.
- **One history-token sum, no filters.** `countable_roles = LLM_ROLES | ROLE_REMAP.keys()`. The budget calc and `history_entry.tokens_estimated` reuse the same value — single source of truth.
- **Explicit `combined` for LLM assembly.** `combined = [*history, *wiki_msgs, *notes_msgs, *memory_msgs, user_msg]` is iterated to build the LLM message list, instead of mutating `history` and iterating it.
- **`ROLE_REMAP` lifted to module-level constant.** It's now referenced from both the LLM-remap loop and the countable-roles set.
- **`user_message` gets its own SourceEntry.** Pre-refactor the role-based filter incidentally counted the current-turn user message; the new role-allowlist filter doesn't (current-turn content isn't in `history` during the calc), so an explicit SourceEntry restores diagnostics parity.

The architectural property worth preserving: any future auto-injection role just needs its own SourceEntry. No filter to keep in sync; no double-counting risk.

## Changes

- `src/decafclaw/context_composer.py` — refactor compose(), lift `ROLE_REMAP` to module-level, add `user_msg_entry` SourceEntry, update docstring.
- `tests/test_context_composer.py` — two new tests in `TestHistoryArchivedRemap`:
  - `test_archived_remap_messages_counted_in_history_entry` — regression test for #393.
  - `test_source_entries_account_for_all_llm_content` — structural no-double-count guard rail with a 5% / 50-token tolerance. The 50-token floor absorbs the self-referential `[Context: ...]` status line that's appended after all SourceEntry computation by design.
- `docs/context-composer.md` — new "History accounting" sub-section under "Token budget".
- `docs/dev-sessions/2026-04-29-1504-token-accounting-archived-remap-393/` — spec / plan / research / notes.

## Test Plan

- [x] `make test` — full suite: 2312 passed, 0 failed.
- [x] `make check` — lint + typecheck clean (ruff + pyright + tsc).
- [x] Real-data smoke: load an archived conversation with 8 auto-injected remap-role messages, call `compose()`, verify `history_entry.tokens_estimated` includes them. Pre-fix: 2623 tokens. Post-fix: 9569 tokens.

## References

- Spec: `docs/dev-sessions/2026-04-29-1504-token-accounting-archived-remap-393/spec.md`
- Plan: `docs/dev-sessions/2026-04-29-1504-token-accounting-archived-remap-393/plan.md`
- Closes #393

### DIFF
--- a/docs/context-composer.md
+++ b/docs/context-composer.md
@@ -88,6 +88,14 @@ compaction prompts.
 - Tool definitions are fixed overhead on every turn — [tool deferral](tool-search.md) keeps this in check
 - Vault context competes for remaining budget after fixed costs (see [Vault Retrieval](#vault-retrieval) below)
 
+#### History accounting
+
+`compose()` treats the input `history` list as read-only during all token-budget and diagnostics calculation. Fresh-this-turn injections (`vault_references`, `conversation_notes`, `vault_retrieval`, and the current user message) are each tracked via their own `SourceEntry` (`wiki_entry`, `notes_entry`, `memory_entry`, `user_msg_entry`). The `history` source entry counts everything archived from prior turns, including messages whose role is in `ROLE_REMAP` (auto-injected role messages archived from earlier turns) — they're sent to the LLM after remap, so they count.
+
+After all token accounting and message-list assembly, `compose()` appends this turn's injections to `history` so the caller (agent loop) sees the post-turn state on subsequent turns. The single mutation point at the end of `compose()` replaces a previous shape that mixed input and scratchpad uses of `history`.
+
+This contract closes a class of underreporting bugs (#393) where a role-based filter excluded archived auto-injected messages even though the LLM still sees them after remap.
+
 ## How it works
 
 ### Lifecycle
--- a/docs/dev-sessions/2026-04-29-1504-token-accounting-archived-remap-393/notes.md
+++ b/docs/dev-sessions/2026-04-29-1504-token-accounting-archived-remap-393/notes.md
@@ -1,0 +1,19 @@
+# Notes — Token accounting for archived role-remap messages
+
+## Session log
+
+- **2026-04-29 15:04** — Session started. Worktree at `.claude/worktrees/fix-token-accounting-393/`. Baseline `make test`: 2243 passed. Issue #393 seeded into `spec.md`.
+- **Phase 1** — `75fb428` adds failing regression test; `6c59e00` fixup adds tool/memory mock context managers to match sibling `TestCompose` pattern (responsive to code review).
+- **Phase 2** — `7207f0b` refactors `compose()`. Phase 1 test goes RED → GREEN. Full suite 2244/2244, `make check` clean. No existing tests required adjustment.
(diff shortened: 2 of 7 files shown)


lmorchard:
Thanks @copilot — both comments are valid. Addressed:

**Comment 2 (structural test clarity):** Fixed. The structural test now explicitly excludes the `tools` source from the `source_total` sum, with a comment explaining that tool definitions go via the separate API `tools=` parameter and aren't in `composed.messages`. Makes the invariant robust to fixture changes that introduce non-empty default tool defs. Squashed into the PR.

**Comment 1 (`background_event` accounting gap):** Filed as #435 and explicitly out of scope here. It's the same shape of bug as #393 but pre-existing (the pre-#434 role filter also excluded `background_event`). The right fix needs a small design call on whether to add it to `countable_roles` or pre-expand the synthetic messages — better as its own focused PR than to expand this one's scope.

PR #496, opened 2026-05-15
feat(agent): iteration budget with grace turn (closes #448)

### LINKED ISSUE #448: Iteration budget with grace turn (replace hard max_tool_iterations)
## Context

`agent.py` enforces `config.agent.max_tool_iterations` as a hard counter — when hit mid-loop, the conversation cuts off with the LLM having no chance to say "I ran out of room, here's what I have so far." Hermes-agent uses an `IterationBudget` with `consume()` / `refund()` semantics plus one "grace turn" when budget hits zero (lets the model emit a final no-tools response).

Surveying hermes-agent during the 2026-05-12 inspiration scan.

## Opportunity

Cleaner end-of-budget UX: the model gets to wrap up instead of getting cut. `refund()` also lets tool-retry logic give back budget when the retry is "free" (e.g., LLM produced an unparseable tool call and we ask again).

## Proposed approach

- New class in `agent.py` or a new module:
  ```python
  @dataclass
  class IterationBudget:
      remaining: int
      _grace_used: bool = False
      def consume(self) -> bool: ...      # returns False when exhausted
      def refund(self) -> None: ...
      def grace_turn(self) -> bool: ...   # True once, then False forever
  ```
- Agent loop checks budget after each iteration; on exhaustion, sets a flag that forces the next LLM call with `tools=[]`, then returns.
- `max_tool_iterations` becomes the budget's initial value — no config change for users.

## Validation

- Existing `tests/test_agent.py` budget-exhaustion paths updated to assert one grace turn fires.
- New test: budget=0 + `grace_turn()` fires exactly once with no tools.
- Smoke: drive a Mattermost session into exhaustion, verify the model says "I'm at my iteration limit, here's where I am" cleanly.

## Open questions

- Should `delegate_task` child agents share or have independent budgets? (Probably independent — already the shape with `max_tool_iterations` per fork.)

## Reference

`hermes-agent/run_agent.py` (`IterationBudget` class + `grace_call` flag in the main loop).

lmorchard:
## Summary
- Replace the hard `for iteration in range(max_tool_iterations)` counter in the agent loop with an `IterationBudget` object that grants one **grace turn** at exhaustion.
- When the budget runs out, the model gets one final no-tools LLM call to wrap up coherently instead of being cut off mid-turn. Cleaner end-of-budget UX, and `refund()` provides a "free retry" hook for cases where the LLM produced nothing usable.

## Design Decisions
- **Standalone module** (`src/decafclaw/iteration_budget.py`) over inline in `agent.py` — matches the pattern of `compaction_decisions.py` / `context_cleanup.py` (small focused modules around the agent loop), and the class is trivially unit-testable in isolation.
- **Grace turn appends a terse system note** ("You have reached your tool iteration budget. Produce a final response…") before the `tools=[]` LLM call — most explicit signal to the model that it has hit the limit. Rejected: rely on the model to infer from an empty tools list.
- **`refund()` wired only into the empty-response retry path.** That retry is unambiguously "free" (the model produced nothing). Reflection retries are substantive critique cycles and stay on the budget. Rejected: ship `refund()` with no callers, or wire into reflection retries too.
- **Two fall-back paths preserve the always-something contract.** If the grace LLM call raises OR returns empty content, the loop falls through to today's `_finalize_max_iterations()` notice + accumulated text. The user always sees output.
- **No user-visible config change.** `max_tool_iterations` still seeds the budget's initial `remaining`. No new env vars, no new fields. Child agents continue to receive their own forked `child_max_tool_iterations` budget via `delegate.py` — child-agent budget semantics unchanged.

## Changes
- `src/decafclaw/iteration_budget.py` — new module with `IterationBudget` dataclass exposing `consume()` / `refund()` / `grace_turn()`.
- `src/decafclaw/agent.py` — `TurnRunner.run()` swaps the `for` loop for `while budget.consume():`. New `_run_grace_turn()` method handles the one-shot grace LLM call with exception + empty-content fall-backs. Empty-response retry path now calls `budget.refund()`. `budget` declared as a `TurnRunner` dataclass field (CLAUDE.md "new runtime state goes on the dataclass").
- `tests/test_iteration_budget.py` — 8 unit tests covering initial state, consume exhaustion, refund restoration, grace_turn one-shot.
- `tests/test_agent_turn.py` — replaced the two existing max-iterations tests with three new ones (grace-turn success, exception fallback, empty-content fallback). Added empty-retry refund test asserting the retry runs as a normal iteration.
- `docs/architecture.md`, `docs/reflection.md` — one-line updates describing the grace turn.

## Test Plan
- [x] `make check` passes
- [x] `make test` passes — 2517 passed (9 new tests vs main baseline)
- [ ] Live smoke: drive a Mattermost or web-UI session into exhaustion (set `MAX_TOOL_ITERATIONS=3` and ask for a research task that needs more rounds), confirm the model produces a clean "I'm at my iteration limit, here's where I am" wrap-up rather than a hard cutoff.

## References
- Spec: `docs/dev-sessions/2026-05-15-1218-iteration-budget-grace-turn/spec.md`
- Plan: `docs/dev-sessions/2026-05-15-1218-iteration-budget-grace-turn/plan.md`
- Closes #448

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -124,7 +124,7 @@ Every event carries enough correlation info (`tool_call_id`, `context_id`) that 
 
 1. **Setup** — fork the context if needed, resolve the active model, apply task mode
 2. **Compose context** — `ContextComposer.compose()` builds the full message array: system prompt, proactive vault retrieval, referenced pages, history, and classifies tools into active vs deferred. Pre-emptive tool search runs a keyword match against the user message to auto-promote relevant deferred tools into the active set for this turn. Returns a `ComposedContext`. See [Context Composer](context-composer.md), [Tool Priority](tool-priority.md), [Pre-emptive Tool Search](preemptive-tool-search.md).
-3. **Iteration loop** — up to `max_tool_iterations` (default 200):
+3. **Iteration loop** — up to `max_tool_iterations` (default 200) tool-call rounds, plus one **grace turn** (a final no-tools LLM call when the budget is exhausted, so the model can summarize where it is instead of being cut off):
    - Build the per-iteration tool list (fetched tools may change mid-turn as the model calls `tool_search`)
    - Call the LLM via the provider abstraction — streaming tokens are published as events
    - If the response has text and no tool calls, break out of the loop
--- a/docs/dev-sessions/2026-05-15-1218-iteration-budget-grace-turn/plan.md
+++ b/docs/dev-sessions/2026-05-15-1218-iteration-budget-grace-turn/plan.md
@@ -1,0 +1,426 @@
+# Iteration Budget with Grace Turn Implementation Plan
+
+**Goal:** Replace the hard `for ... range(max_tool_iterations)` counter in `agent.py` with an `IterationBudget` object that grants one grace turn at exhaustion, plus a `refund()` path for the empty-response retry.
+
+**Approach:** Add a standalone `iteration_budget.py` module with the class (Phase 1). Wire it into the agent loop, adding a no-tools LLM call ("grace turn") when budget runs out and a fall-back path when the grace call errors (Phase 2). Wire `refund()` into the empty-response retry path so it doesn't eat budget (Phase 3). Update the two doc pages that describe the iteration cap (Phase 4).
+
+**Tech stack:** Python 3.x, `dataclasses`, `asyncio`, `pytest` + `pytest-asyncio`.
+
+---
+
+## Phase 1: `IterationBudget` module + unit tests
+
+Adds a standalone, framework-free class with `consume()`, `refund()`, and `grace_turn()` semantics. No callers yet — Phase 2 wires it in. Independently valuable: the class is fully unit-tested and could be used anywhere a budget pattern is needed.
+
+**Files:**
+- Create: `src/decafclaw/iteration_budget.py`
+- Create: `tests/test_iteration_budget.py`
+
+**Key changes:**
+
+`src/decafclaw/iteration_budget.py`:
+
+```python
+"""Iteration budget for the agent loop with one grace turn at exhaustion."""
+
+from __future__ import annotations
+from dataclasses import dataclass
+
+
+@dataclass
+class IterationBudget:
+    """Tracks remaining tool-call iterations with one-shot grace-turn semantics.
+
+    The agent loop calls ``consume()`` at the top of each iteration. When it
+    returns False, the budget is exhausted; the loop then calls
+    ``grace_turn()`` once to gate a single no-tools final LLM call.
+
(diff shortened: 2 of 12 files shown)


lmorchard:
Copilot review addressed in https://github.com/lmorchard/decafclaw/pull/496/commits/8e2244e:

- **Grace turn now bumps `ctx._current_iteration`** before the LLM call so `llm_start` / `llm_end` events don't reuse the last tool-enabled iteration's number. Event consumers (UI iteration counter, diagnostics sidecar) now see a fresh iteration for the grace call.
- Added `test_run_agent_turn_grace_turn_bumps_current_iteration` that captures `ctx._current_iteration` at each LLM call and asserts the grace call uses a strictly-higher number than the last normal iteration.

lmorchard:
Iterated on the grace prompt and verified end-to-end via a real-LLM eval:

**Smoke-test finding:** the previous prompt ("Produce a final response summarizing your progress; you cannot call any more tools") wasn't directive enough — models kept trying to do the original task and bailed mid-stream when they realized they couldn't call a tool (exactly the failure mode Les hit during manual smoke).

**Prompt change:** switched to a user-role nudge with explicit `STOP` / `Do not continue` framing, matching the reflection-critique pattern at agent.py:800-810. User-role weights heavier than system mid-turn.

**Eval framework extension:** `eval/runner.py:_run_one` now honors `setup.max_tool_iterations` so the eval can self-describe its budget. Adds `evals/grace-turn.yaml` — drives a sequential shell-tool prompt with budget=3 and asserts: (a) wrap-up names the limit, (b) fallback notice absent, (c) ≤3 tool calls. Passes on `vertex-gemini-flash`. Sample grace-turn output:

> "I have executed the `date`, `echo \/Users/lorchard`, and `uname -s` commands […]. Unfortunately, I have reached my tool iteration limit and cannot execute the remaining commands or provide the final summary."

Docs updated (`docs/eval-loop.md` for the new setup field). `make test` 2518 pass, `make check` clean.

lmorchard:
Copilot review round 2 addressed in 960e469:

- **agent.py:_run_grace_turn docstring** — updated from "appends a terse system note" to "appends a directive user-role nudge", with a brief why-note about user-role weighting.
- **spec.md design decision** — rewrote the "Grace turn appends a terse, action-only system note" line to reflect the user-role decision with the smoke-test rationale. Added an explicit "system-role: rejected" entry.
- **spec.md Patterns to follow** — updated the message-role example from `role: "system"` to `role: "user"` with the `[iteration_limit]` tag, and pointed at the reflection-critique pattern as precedent.
- **plan.md** — left the original draft code block in place as historical record but flagged inline that the role and wording were iterated post-implementation. Same note added to the manual-verification checkbox.

PR #506, opened 2026-05-15
fix(cancel): persist cancel marker to archive (#491)

### LINKED ISSUE #491: Cancelled turns persisted as `[cancelled]` cause the agent to re-fulfill the request on resume
## Context

Surfaced during live smoke of the TUI spike (#489).

Sequence:

1. User: "write me a 600-word essay about anything at all"
2. Agent: starts streaming an essay...
3. User: Ctrl+C (cancel)
4. Server emits \`message_complete\` with \`text: "[cancelled]"\` (conversation_manager.py:1080); the partial assistant content is dropped.
5. WS reconnect happens (this part is unrelated but happens to be when I observed it)
6. User: "hello"
7. Agent: writes a *fresh* full 600-word essay on a different topic, *then* responds "Hello! How can I help you today?"

## Diagnosis

The conversation archive after the cancel looks like:

```
user:      write me a 600-word essay about anything at all
assistant: [cancelled]
user:      hello
```

The LLM sees an open request from the user and a literal "[cancelled]" string from the assistant — which doesn't carry strong semantic weight. The LLM apparently interprets this as "I never delivered what was asked; let me do it now," and tackles both requests in sequence.

## Suggested directions

- **Persist the partial assistant content** as the assistant message, with cancellation expressed as a separate system/marker line. The LLM would see `user(essay) → assistant(<partial>) → user(hello)` and likely treat the partial as a final-enough answer.
- **Stronger cancel marker**: persist the message as `[user cancelled this request]` or as a system-role entry, so the LLM has explicit signal that this turn is closed and shouldn't be retried.
- **Inject cancel context into the next system prompt** when the most-recent assistant message is a cancel sentinel.

## Repro

Trigger via TUI (#489 once merged) or web UI: start a long-streaming response, cancel mid-stream, then send any new message in the same conversation.

## References

- Cancel emission: \`src/decafclaw/conversation_manager.py:1080\`
- TUI cancel-preserve fix (client-side display only; doesn't affect archive): commit \`567cc8e\` on \`tui-spike\`
- Surfaced via TUI spec acceptance criterion "Kill make dev, restart, see TUI reconnect and resume the same conversation"

## Suggested fields

- Priority: P2
- Size: S/M (depends on direction — partial-content persistence may need archive schema thought)

lmorchard:
## Summary
- Cancelling a streaming agent turn left the conversation archive in a state the LLM read as "user request still open" — the next user message would cause the agent to re-fulfill the cancelled request before responding.
- This PR writes a strong, role-remapped cancel marker to the archive on every cancel path, optionally preceded by any partial assistant text that was streamed before the cancel.

## Design decisions
- **New archive role `cancel_marker`**, added to `context_composer.ROLE_REMAP` → `user`. We can't use `system` mid-conversation: Vertex/Gemini's request builder collapses all `role: system` messages into the top-level `systemInstruction`, losing the "between two user turns" position.
- **Preserve partial streamed content** as an `assistant` row before the marker, when present. The TUI already preserves partial content client-side (commit `567cc8e` on `tui-spike`); persisting it server-side matches that intent and gives the LLM a "I did start answering, then was interrupted" signal.
- **Write-once latch** (`ConversationState.cancel_marker_written`) funnels both the `CancelledError` handler and the manager's normal-completion path through the same archive write. Required because `_check_cancelled` at iteration start returns `_Final` cleanly without raising — `task.cancel()` may not have fired by the time the agent returns.

## Changes
- `src/decafclaw/context_composer.py` — register `cancel_marker` in `ROLE_REMAP`.
- `src/decafclaw/conversation_manager.py` — `CANCEL_MARKER_TEXT` constant, `_write_cancel_archive` helper, `_write_cancel_marker_once` latched method, `note_partial_assistant_archived` API, three new `ConversationState` fields, partial-text accumulation in `on_stream_chunk`, marker write on both completion paths.
- `src/decafclaw/agent.py` — `_archive` notifies the manager when an assistant body lands; `_check_cancelled` keeps the in-history append but no longer archives; `_handle_no_tool_calls` short-circuits on cancel+empty.
- Tests: unit tests for the helper, three end-to-end cancel-shape tests, one composer remap test, one race-coverage test for the iteration-start path. Existing `test_check_cancelled_returns_result_when_cancelled` updated for the new archive-write contract.
- Docs: new "Cancelled turns" subsection in `docs/conversations.md`, cross-link from `docs/context-composer.md`.

## Test plan
- [x] `make test` — 2516 passed (+9 vs baseline)
- [x] `make check` — clean (lint + pyright + tsc + message-types drift)
- [ ] **Live smoke (post-merge):** start a long streaming response in the web UI, cancel mid-stream, send "hello" — agent should NOT regenerate the cancelled essay.
- [ ] **Live smoke (post-merge):** repeat in TUI once #489 lands.
- [ ] **Live smoke (post-merge):** verify the cancel_marker row renders sensibly (or invisibly — both acceptable) in both UIs.
- [ ] **Live smoke (post-merge):** reload page after a cancel, confirm archive replay still shows partial + marker and no LLM retry.

## References
- Spec: `docs/dev-sessions/2026-05-15-1219-fix-491-cancelled-turn-archive/spec.md`
- Plan: `docs/dev-sessions/2026-05-15-1219-fix-491-cancelled-turn-archive/plan.md`
- Notes: `docs/dev-sessions/2026-05-15-1219-fix-491-cancelled-turn-archive/notes.md` (includes deferred follow-ups: same bug shape exists for the generic-exception handler at `conversation_manager.py:1160-1167` and likely for max-iterations exhaustion — both out of scope for this PR)

Closes #491

### DIFF
--- a/docs/context-composer.md
+++ b/docs/context-composer.md
@@ -116,6 +116,8 @@ the slice guidance still lives in the system prompt via the addendum.
 
 `compose()` treats the input `history` list as read-only during all token-budget and diagnostics calculation. Fresh-this-turn injections (`vault_references`, `conversation_notes`, `vault_retrieval`, and the current user message) are each tracked via their own `SourceEntry` (`wiki_entry`, `notes_entry`, `memory_entry`, `user_msg_entry`). The `history` source entry counts everything archived from prior turns, including messages whose role is in `ROLE_REMAP` (auto-injected role messages archived from earlier turns) — they're sent to the LLM after remap, so they count.
 
+`ROLE_REMAP` also covers `cancel_marker` (remapped to `user`), an archive-only role written when a user cancels a turn. See [Cancelled turns](conversations.md#cancelled-turns) for the producer-side details — the remap is what guarantees the marker reaches the LLM at the right position regardless of provider (Vertex collapses `system`-role messages into `systemInstruction`).
+
 After all token accounting and message-list assembly, `compose()` appends this turn's injections to `history` so the caller (agent loop) sees the post-turn state on subsequent turns. The single mutation point at the end of `compose()` replaces a previous shape that mixed input and scratchpad uses of `history`.
 
 This contract closes a class of underreporting bugs (#393) where a role-based filter excluded archived auto-injected messages even though the LLM still sees them after remap.
--- a/docs/conversations.md
+++ b/docs/conversations.md
@@ -22,6 +22,21 @@ Each line is a JSON object with `role`, `content`, optional `tool_calls`/`tool_c
 
 When DecafClaw restarts, it reads the archive for any conversation that receives a new message. The full history is replayed into memory, so the agent picks up where it left off.
 
+### Cancelled turns
+
+When a user cancels an in-progress agent turn (Ctrl+C in the TUI, the cancel button in the web UI), `ConversationManager` persists:
+
+1. Any partial assistant text that was streamed before the cancel (as an `assistant` row). Omitted if no content streamed.
+2. A `cancel_marker` row carrying `CANCEL_MARKER_TEXT`: `"[User cancelled this turn. Do not retry the cancelled request unless they explicitly ask for it again.]"`.
+
+The write happens on **whichever cancel path fires first** — either the `CancelledError` handler when `task.cancel()` propagates an exception, or the normal-completion path when the agent loop observes the cancel cleanly (via `_check_cancelled` at iteration start, or the streaming short-circuit in `_handle_no_tool_calls`). The agent reports the clean-observation via `ConversationManager.note_cancel_observed_by_agent`. A write-once latch (`ConversationState.cancel_marker_written`) makes both entries safe when both fire.
+
+The normal-path write is gated on the agent's explicit `cancel_observed_by_agent` flag rather than the raw `cancel_event` state — so a cancel signal that arrives *after* the agent has already returned a real response doesn't spuriously mark a delivered response as cancelled.
+
+`cancel_marker` is an archive-only role; the [context composer](context-composer.md) remaps it to `user` before sending history to the LLM (matching the `vault_retrieval` / `conversation_notes` pattern, see `ROLE_REMAP`). The marker sits between the cancelled-user-turn and the next user turn at the right position across all providers — Gemini/Vertex collapses `system`-role messages into the top-level `systemInstruction` field, which would lose the positional meaning.
+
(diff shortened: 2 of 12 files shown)


lmorchard:
Addressed Copilot's 5 review comments in 57872de (force-pushed):

1. **agent.py:69 — `partial_assistant_archived` set even on archive failure.** Fixed: track `wrote` inside the try/except and only call `note_partial_assistant_archived` when `append_message` succeeded.

2. **conversation_manager.py:1154 — `cancel_event.is_set()` spuriously triggers marker write for late cancels.** Fixed: introduced `state.cancel_observed_by_agent` set explicitly by the agent (via `note_cancel_observed_by_agent`) when it observes cancel cleanly via `_check_cancelled` or the `_handle_no_tool_calls` short-circuit. Normal-path marker write now gates on this flag rather than the raw event, so a cancel arriving after a real response was delivered no longer marks it as cancelled. New regression test: `test_late_cancel_after_completion_no_marker`.

3. **conversation_manager.py:1081 — O(n²) string concatenation, partial held in memory between turns.** Fixed: replaced `partial_assistant_text: str` with `partial_assistant_chunks: list[str]`; chunks appended in the streaming callback, joined lazily in `_write_cancel_marker_once`. Reset (rebound to `[]`) at turn start releases memory between turns.

4. **conversation_manager.py:709 — partial write succeeds, marker write fails → retry duplicates partial.** Fixed by inlining the two-step write into `_write_cancel_marker_once` and updating `state.partial_assistant_archived = True` immediately after a successful partial append, before the marker write. If the marker append then fails, the latch stays unset (so a retry can write the marker) but the partial flag is set (so the retry skips the partial — no duplicate). Did NOT reverse write order to marker-first: chronological order is what the LLM reads, and `user → cancel_marker → assistant` would scramble the conversation semantically.

5. **docs/conversations.md:33 — doc said "CancelledError handler" but implementation also handles clean cancel.** Fixed: rewrote the subsection to describe both entry paths, the `note_cancel_observed_by_agent` API, and the late-cancel gating.

`make test` clean at 2517 passed (+10 from baseline), `make check` clean.

lmorchard:
Filed follow-up #517 covering items 1 + 2 from the deferred list (generic exception handler + max-iterations/circuit-breaker missing archive signal — same conceptual shape as this PR). Items 3 (dead preempt_search filter) and 4 (UI affordance for cancel_marker) left unfiled — trivial cleanup / depends on post-merge live-smoke.

Issue #486, opened 2026-05-13
ConversationManager: recovery dispatch identity check loses race with fresh request_confirmation


lmorchard:
## Context

Follow-up to PR #484 / issue #440. Surfaced in Copilot review pass #5 on commit \`665b77c\` (comment ID 3236326877). Narrow race that #484 didn't intend to cover.

## The race

In \`src/decafclaw/conversation_manager.py\` \`respond_to_confirmation\`, the recovered-confirmation path (the \`else\` branch where \`waiter_event is None\`):

\`\`\`python
async with state.lock:
    claimed_request = state.pending_confirmation
    state.confirmation_response = response   # claim under lock
    waiter_event = state.confirmation_event

# archive + emit OUTSIDE lock (durability path)
append_message(...)
await self.emit(...)

# (recovery branch — no running waiter)
async with state.lock:
    if state.confirmation_response is not response:  # <-- L527
        log.info(\"Recovery superseded by cancel\")
        return
    state.pending_confirmation = None
    state.confirmation_response = None
\`\`\`

The L527 identity check is supposed to detect \"did a concurrent cancel clear my claim while archive was in flight.\" \`cancel_pending_confirmation\` does check \`state.confirmation_response is not None\` and bails (lines ~569), so cancel-vs-respond is correctly handled.

But a **fresh \`request_confirmation\` for a NEW turn** does NOT check for an existing claim. It unconditionally writes:

\`\`\`python
state.pending_confirmation = request
state.confirmation_response = None   # L772 — resets unconditionally
\`\`\`

## Scenario

1. Old turn issues \`request_confirmation\`, blocks waiting on event.
2. Bot restarts (deploy / crash). \`pending_confirmation\` is in archive, recovered to state on startup. No waiter (\`waiter_event is None\`).
3. User clicks Approve on the OLD confirmation via HTTP / web UI. \`respond_to_confirmation\` runs:
   a. Claims under lock at L472 (sets \`confirmation_response\`).
   b. Releases lock. Archive write at L483 succeeds.
   c. Emit at L492 succeeds.
4. **Between L498 (lock released) and L517 (lock re-acquired)** the user submits a new message via WebSocket. A new turn starts, the agent loop runs and itself calls \`request_confirmation\`. L771-772 fires: \`pending_confirmation = new_request\`, \`confirmation_response = None\`.
[...]

PR #513, opened 2026-05-15
fix(conversation_manager): queue concurrent request_confirmation calls per conv (closes #485)

### LINKED ISSUE #485: request_confirmation overwrites concurrent in-flight confirmation slot
**Pre-existing issue surfaced during #484 (Copilot review on `request_confirmation` line 781).**

## Problem

`ConversationManager.request_confirmation` unconditionally writes `state.pending_confirmation` / `state.confirmation_event` / `state.confirmation_response` when called, even if there's already a live confirmation in flight on the same conv. The lock from #484 makes this transition atomic, but doesn't prevent the overwrite itself.

Two tool calls executed concurrently via `execute_tool_calls` (which creates one task per tool call) that BOTH request confirmation will clobber each other's slot — the second's `pending_confirmation = request` overwrites the first's, leaving the first waiter parked on its own captured event reference while the manager's state belongs to the second confirmation. The first waiter then times out or hangs while its archived request remains unresolved.

## Why this was out of scope for #484

The PR scope (per `docs/dev-sessions/2026-05-13-0939-conv-manager-lock/spec.md`) is replacing boolean concurrency flags with `asyncio.Lock` and closing the read-yield-write race in the dispatch / confirmation lifecycle. The overwrite behavior is a design question about how multiple in-flight confirmations on a single conv should be handled — predates #440 and would expand the spec significantly.

## Options

1. **Reject** — return a denial response immediately if `state.pending_confirmation is not None`. Simple; tool authors need to handle "confirmation rejected because another is pending."
2. **Queue** — serialize confirmation requests per conv via an asyncio.Queue or a per-conv `confirmation_queue: list[ConfirmationRequest]`. More complex but transparent to tool authors.
3. **Block / wait** — second `request_confirmation` awaits the first to complete before claiming the slot. Simple but increases turn latency under contention.

## Reproduction

In current code (post-#484): call `request_confirmation` twice concurrently on the same `conv_id`. The first waiter's parked event will never be signaled because the second call overwrote `state.confirmation_event` with a new Event object before the first's responder could `event.set()` on the old one.

## Refs

- PR #484 — `asyncio.Lock` migration that made this overwrite atomic but didn't address the design issue
- Copilot review comment on `request_confirmation` (line 781)

lmorchard:
## Summary

- Replace ``request_confirmation``'s unconditional pending-slot overwrite with a per-conversation FIFO queue. Concurrent approval requests on the same conversation (common with delegated background tasks producing parallel prompts) are now serialized — the user sees them one at a time, each waits for its own response.
- Why: the pre-#484 race made the second concurrent call clobber the first's ``pending_confirmation`` / ``confirmation_event`` / ``confirmation_response`` triple. The lock from #484 made the transition atomic but didn't stop the overwrite, leaving the first waiter parked on a dead event indefinitely.

## Design Decisions

- **Queue, not reject or block.** Background-task workloads legitimately produce parallel approval prompts; rejecting would silently deny tools that actually need user approval, and blocking would tie up turn latency under contention. Queueing is transparent to callers.
- **Active slot stays canonical.** ``state.pending_confirmation`` / ``confirmation_event`` / ``confirmation_response`` keep their meaning. Queue is the list of *not-yet-active* requests. Existing code paths (respond, cancel, startup recovery, archive) keep working on the active slot.
- **Per-entry event.** Each ``_QueuedConfirmation`` carries its own ``asyncio.Event``. A single ``_promote_next_confirmation_unlocked`` helper pops the head, installs it in the active slot, signals its event, and is called from every clear-active path.
- **Archive at enqueue, emit at promote.** Crash recovery sees every requested confirmation in submission order; UI sees exactly one active confirmation at a time.
- **Timeout starts at promote.** ``request.timeout`` is the post-promote countdown. While queued, the waiter blocks on ``promoted_event`` with no timeout — a long-running active confirmation can't silently deny everything queued behind it.
- **#486 stays separate.** Queueing reduces but does not eliminate the recovery-dispatch L527 identity-check race window. Tracked there, fixed separately.

## Changes

- ``src/decafclaw/conversation_manager.py``:
  - New module-level ``_QueuedConfirmation`` dataclass.
  - New module-level ``_confirmation_request_payload`` helper (the existing emit payload, shared across emit sites).
  - New ``confirmation_queue: list[_QueuedConfirmation]`` field on ``ConversationState``.
  - New ``_promote_next_confirmation_unlocked`` method on ``ConversationManager`` with a defensive precondition that returns ``None`` if the active slot is still claimed.
  - ``request_confirmation`` restructured into active vs. queued paths. The queued path blocks on ``promoted_event`` without a timeout, then runs the existing post-wait claim block. Cancellation cleanup handles both still-queued and just-promoted shapes.
  - ``respond_to_confirmation``'s recovery branch and ``cancel_pending_confirmation``'s no-waiter branch each call promote after their cleanup so queued items behind a recovered or unsignaled active confirmation aren't orphaned.
  - Lock-contract comment on ``ConversationState.lock`` extended to cover queue mutations.
- ``tests/test_conversation_manager.py``: 7 new tests under the "Confirmation queue (issue #485)" section.
- ``docs/architecture.md``: one-line note in the Confirmations section.

## Test Plan

- [x] ``make check`` (lint, typecheck, JS check)
- [x] ``make test`` (2514 passed, +7 new)
- [ ] Live test in web UI: trigger two delegated background tasks that each request approval; confirm both prompts surface serially, each gets its own response.

## References

- Spec: ``docs/dev-sessions/2026-05-15-1218-fix-485-request-confirmation-overwrite/spec.md``
- Plan: ``docs/dev-sessions/2026-05-15-1218-fix-485-request-confirmation-overwrite/plan.md``
- Closes #485

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -162,6 +162,8 @@ The rule: **one agent turn per conversation at a time, unlimited concurrent conv
 
 When a tool needs user approval (shell commands, skill activation), the ConversationManager persists the confirmation request as a JSONL archive entry with `role: "confirmation_request"`. The agent loop suspends mechanically until it receives a matching `confirmation_response` entry. Pending confirmations survive page reload and server restart — a startup scan recovers them. See `confirmations.py` for the `ConfirmationAction` enum and handler registry.
 
+Concurrent `request_confirmation` calls on the same conversation (common with delegated background tasks producing parallel approval prompts) are serialized through a per-conversation FIFO queue: the user sees them one at a time, each waits for its own response, and `request.timeout` starts only when a queued request becomes active.
+
 ## Transport adapters
 
 Each transport is responsible for:
--- a/docs/dev-sessions/2026-05-15-1218-fix-485-request-confirmation-overwrite/plan.md
+++ b/docs/dev-sessions/2026-05-15-1218-fix-485-request-confirmation-overwrite/plan.md
@@ -1,0 +1,321 @@
+# Queue concurrent request_confirmation — Implementation Plan
+
+**Goal:** Replace `request_confirmation`'s unconditional pending-slot overwrite with a per-conversation FIFO queue so simultaneous approval requests (esp. from delegated background tasks) all reach the user serially.
+
+**Approach:** Add a `confirmation_queue: list[_QueuedConfirmation]` on `ConversationState`. When `request_confirmation` lands while `state.pending_confirmation is not None`, enqueue a `_QueuedConfirmation(request, event)` and block on its own event until promoted. A single `_promote_next_confirmation_unlocked` helper, called from every clear-active path (respond, cancel, timeout, cancel-raced), pops the next waiter and signals it. Timeout starts at promote time, not at enqueue. #486 stays separate.
+
+**Tech stack:** Python 3, `asyncio.Lock` / `asyncio.Event`, existing `confirmations.py` types, pytest + pytest-asyncio. No new dependencies.
+
+---
+
+## Phase 1: Queue + concurrent-request regression slice
+
+End-to-end slice: failing regression test → queue scaffolding → enqueue / promote / wire-up at all clear-active call sites → test passes. After this phase, two concurrent `request_confirmation` calls on the same conv resolve to their own responses in submission order.
+
+**Files:**
+- Modify: `src/decafclaw/conversation_manager.py`
+  - Add `_QueuedConfirmation` dataclass (module-level, near `ConversationState`).
+  - Add `confirmation_queue: list[_QueuedConfirmation] = field(default_factory=list)` to `ConversationState` (next to the existing confirmation-state triple at L188–191).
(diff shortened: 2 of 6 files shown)


lmorchard:
Addressed Copilot review feedback in the latest squash:

1. **Race fix (line 976):** ``_QueuedConfirmation`` now carries an ``active_event`` field; the promote helper writes it before signalling ``promoted_event``, and the queued waiter captures it directly instead of re-reading ``state.confirmation_event`` after wake. Closes the narrow race window Copilot described.
2. **Dead assignment (line 708):** ``cancel_pending_confirmation``'s no-waiter branch no longer sets ``confirmation_response`` only to immediately clear it. The waiter-present branch still claims the slot as before.
3. **Shared cleanup helper (line 1050):** Extracted ``_clear_active_and_promote_unlocked``. The three post-wait branches and the queued-waiter cancellation path all call it, so future branches inherit the clear+promote sequence by construction.
4. **Stale comment (line 560):** Updated to describe both waiter-present and no-waiter behavior, including the explicit promote call added in the recovery branch.

``make check`` + ``make test`` still green (2514 passed).

PR #518, opened 2026-05-16
feat(skills): always-loaded code_execution sandbox (closes #471)

### LINKED ISSUE #471: Programmatic tool calling: LLM-written Python sandbox calling decafclaw tools
## Context

The most architecturally novel pattern in hermes-agent. When the LLM needs to do something that's "multiple tool calls + glue logic" (read 5 vault pages → extract dates → return earliest), today it requires 12 round-trips through the conversation, all of which land in context. Hermes-agent's `code_execution_tool.py` lets the LLM write a Python script that calls decafclaw tools over an RPC channel, and only the final result lands in the conversation. Intermediate results stay off-context.

Big win for context efficiency on deterministic multi-step work. Also a power surface for "let me actually compute this."

## Proposed approach

- New `src/decafclaw/tools/code_execution.py`.
- Spawn subprocess with restricted Python; expose a `dc.*` proxy that calls back into the main process via UDS (Unix initially).
- Sandbox-accessible tools are a curated subset (read-only by default — `vault_read`, `workspace_read`, `http_get`); write tools opt-in via sandbox config.
- Timeout + memory cap on the subprocess.
- Return: stdout, stderr, return value, time taken, list of tools called with arg summaries.

## Validation

- Test: simple script that calls `dc.vault_read("Foo")` returns expected content.
- Test: timeout fires when the script loops forever.
- Test: write tool blocked when sandbox is read-only.
- Security review: subprocess isolation, env-var passthrough, FD leaks, signal handling.

## Open questions

- Read-only by default vs. allowlist write tools? Probably read-only with explicit opt-in per skill.
- macOS / Linux initially; Windows later if needed.
- Worth a brainstorm before committing — this touches the security model meaningfully.

## Reference

`hermes-agent/tools/code_execution_tool.py`.

lmorchard:
## Summary

- Add a bundled always-loaded `code_execution` skill that runs an LLM-authored Python script in a subprocess and dispatches a curated 11-tool subset of decafclaw tools over a Unix-domain-socket RPC channel.
- Intermediate per-tool outputs stay off the conversation transcript — only the script's final stdout/stderr/return lands in the result. Targets deterministic multi-step work where per-step context is wasted (e.g. "read 5 vault pages, extract a field from each, return the earliest").

## Design Decisions

- **Bundled always-loaded skill**, not on-demand. Broadly useful; the `SkillConfig` abstraction gives us per-instance cap overrides without inflating top-level `Config`. Tool surface inflates by exactly one.
- **Curated allowlist is the trust boundary.** 11 tools that don't require user confirmation. The subprocess can't pause mid-execution for a prompt, so confirmation-gated tools (`shell`, `send_email`, `http_request`) are excluded by design. Tools that *sometimes* gate (e.g. `vault_write` outside `agent/`) are handled by forking ctx with `request_confirmation=None`, which routes them to their existing non-interactive error paths.
- **Subprocess + UDS + JSON-line framing**, stdlib only. Cleanly killable, memory-cappable via `RLIMIT_AS`, no extra deps. macOS gets no `RLIMIT_AS` enforcement (kernel rejects lowering below `RLIM_INFINITY`); gated behind `sys.platform.startswith("linux")` at module load. Phase 4's memory test skips on darwin.
- **Defense in depth.** Both the script-side stub (which only generates accessors for allowlisted tools) and the server-side RPC handler check the allowlist against the same `SANDBOX_ALLOWED_TOOLS` tuple.
- **Resource caps match hermes-agent's reference impl**: 300s wall-clock, 50 RPC calls, 50KB stdout / 10KB stderr (head 40% + tail 60% truncation), 512MB `RLIMIT_AS`. All overridable via `SkillConfig`.
- **Env scrubbing**: safe-prefix passthrough (`PATH`, `HOME`, `USER`, `LANG`, `LC_*`, `TERM`, `TMPDIR`, `PYTHONPATH`, `VIRTUAL_ENV`, `DECAFCLAW_*`); deny any var whose name contains `KEY`/`TOKEN`/`SECRET`/`PASSWORD`/`CREDENTIAL`/`PASSWD`/`AUTH` (deny takes precedence over passthrough).

## Changes

- **New bundled skill** `src/decafclaw/skills/code_execution/{SKILL.md, tools.py, _sandbox.py, _stub.py}`. The skill is always-loaded; its single `code_execution` tool has `timeout: None` so the internal 300s cap is authoritative.
- **`_sandbox.run_script`** owns subprocess + UDS-server lifecycle, env scrubbing, `RLIMIT_AS` preexec, timeout-with-grace, head+tail truncation, and tempdir cleanup. Phase 4 added a `progress_publish` keyword so per-RPC `tool_status` events flow to transports for inline progress display.
- **`_make_tool_handler`** forks ctx and dispatches via `decafclaw.tools.execute_tool` against the 11-tool allowlist. Wraps the dispatch in `try/except` because Phase 2's `_serve_rpc` deliberately doesn't catch handler exceptions (so handler-level errors would otherwise hang the subprocess).
- **29 tests** across 4 files (skill loading, sandbox plumbing, tool dispatch, resource caps + progress events). Memory-cap test marks `skipif(sys.platform == "darwin")`.
- **Docs**: new `docs/code-execution.md` (overview, allowlist, caps, security boundary, "adding tools to the allowlist" checklist). Updated `docs/skills.md`, `docs/tools.md`, `docs/index.md`, `docs/context-composer.md`. CLAUDE.md `Skills (bundled)` line updated.
- **Eval**: 4 disambiguation cases at `evals/tool_choice/code_execution.yaml`. The three anti-overuse cases pass; the positive case fails because the LLM picks `vault_list` first as discovery — `vault_list` is in `near_miss` so the eval signal is honest, but the description quality doesn't warrant tightening (exploratory decomposition is defensible behavior).

## Test Plan

- [x] `make lint`, `make typecheck`, `make check`, `make test` (2552 passed, 1 skipped on darwin)
- [x] `pytest tests/skills/test_code_execution_*.py -v --durations=10` (slowest test 0.57s, well under 5s budget; no tempdir or zombie-process leaks after suite)
- [x] `make eval-tools` (3 of 4 new cases pass; pre-existing failures unrelated)
- [ ] **Manual smoke in dev** — Phase 6 of the dev session:
  - Single-page lookup picks `vault_read`, not `code_execution`
  - Multi-page synthesis picks `code_execution`, script does N `dc.*` calls
  - Conversation transcript stays clean of intermediates — only the final summary appears
  - Three `tool_status` progress lines appear inline during the call
  - Force timeout: result returns `status: timeout` within ~302s
  - Force allowlist rejection (`dc.shell(...)`): result returns `.error`; no shell confirmation appears in UI
  - No `/tmp/dc-codeexec-*` directories left after a few runs
  - Resume across web UI reload mid-run

## References

- Spec: `docs/dev-sessions/2026-05-15-1244-programmatic-tool-calling-471/spec.md`
- Plan: `docs/dev-sessions/2026-05-15-1244-programmatic-tool-calling-471/plan.md`
- Session notes (flags `delegate.py:394` follow-up + Context.fork concurrent-subobject sharing for future sessions): `docs/dev-sessions/2026-05-15-1244-programmatic-tool-calling-471/notes.md`
- Reference implementation studied: hermes-agent's `tools/code_execution_tool.py`
- Closes #471


lmorchard:
## Parking this PR

Closing without merge after the second Copilot review surfaced an architectural framing-vs-reality gap. The "sandbox" is a subprocess runner with stdlib unrestricted — a script can `import subprocess; subprocess.run([...])` and bypass the shell-confirmation gate, `import urllib.request` and bypass the `http_request` gate, `import os; os.remove(...)` and bypass workspace-write guards. The `dc.*` allowlist is the *advertised* surface but not the *actual* trust boundary.

The "we trust the agent's intent" threat model that this design implicitly assumes is too thin in practice — prompt injection arriving via vault retrieval, notes, or archived conversation content can derail the LLM (we saw this concretely during smoke testing). A real isolation primitive (WASM-based runtime or similar capability boundary) is the architecturally honest answer.

## Where the work went

- **Orthogonal improvements salvaged → #537** ("populate `ToolResult.data` on read-heavy tools"). The structured `.data` shapes on `vault_*`, `workspace_*`, `notes_*`, `tabstack_*` benefit every LLM-direct caller, not just the parked sandbox. That PR is ready to land independently.
- **Architectural replacement tracked → #538** ("code_execution: WASM-based runtime to replace parked subprocess sandbox"). Includes the open design questions (runtime choice, host-import bridge, resource caps), motivating use cases, and references back to the salvageable design patterns in this PR.

## What's preserved here for reference

This branch (`programmatic-tool-calling-471`) stays available. The patterns worth re-using in the WASM redesign:

- `_make_tool_handler` dispatch shape (fork ctx, suppress confirmation, route via `execute_tool`)
- The curated 13-tool allowlist (with the addition of `workspace_glob` / `workspace_search` from the late-session iteration)
- `_render_result` text+widget+data pattern for the script body
- Auto-inject ergonomics for the `dc` proxy import
- Tests pinning `.data` shape per allowlisted tool (already lifted to #537)

## Outstanding follow-ups from this session

- #519 — Copy conversation as JSONL / markdown (web UI)
- #522 — `delegate_task` / `delegate_tasks` in the sandbox (motivating use case: dev-sessions audit)
- #523 — Extend `.data` to non-sandbox tools (broader rollout)
- #537 — Salvaged `.data` work (this session)
- #538 — WASM-based code_execution (replaces this PR)

Issue #255, opened 2026-04-10
Design: first-class workflow abstraction (LangGraph-inspired)


lmorchard:
## Context

The project skill implements a workflow (brainstorm → spec → plan → execute → done) using a state machine + dynamic tool loading + end_turn signals. This works but the workflow mechanics are spread across multiple layers:

- State machine in `skills/project/state.py`
- Tool gating via `get_tools(ctx)` in `tools.py`
- Turn control via `end_turn` / `EndTurnConfirm` on `ToolResult`
- Confirmation gates via event bus in `agent.py`

## Proposal

Design a first-class workflow abstraction inspired by LangGraph's graph-based approach. A workflow would be a declarative structure that:

- Defines **phases/nodes** with allowed tools per phase
- Defines **transitions** (edges) with conditions and gates (confirmation, input, etc.)
- Handles **turn control** automatically (phase boundaries end/pause turns)
- Provides **state persistence** for resumability across conversations
- Is **composable** — skills could define workflows without knowing about agent loop internals

This would replace the ad-hoc layering of state machine + dynamic tools + end_turn that the project skill currently uses, and make it available to any skill that needs structured multi-step workflows.

## Key design questions

- Should workflows be defined in code (Python DSL) or data (YAML/JSON)?
- How do workflows interact with the existing skill system?
- How does the agent loop hand off control to a workflow?
- How do workflows compose with each other (nested workflows)?
- What's the relationship between workflow state and conversation state?

## References

- LangGraph: graph-based agent workflows with mechanical control flow
- Anthropic "Building Effective Agents": workflow vs agent distinction
- CoALA: cognitive architectures for language agents (environment-controlled loops)
- Current implementation: PR #232 (spec-plan-execute-skill)

PR #557, opened 2026-05-20
feat(workflow): first-class workflow engine, conversation-scoped (#255)

### LINKED ISSUE #561: Workflow tools: decide on critical vs normal priority
Follow-up from #557 (workflow engine PR).

The spec [docs/dev-sessions/2026-05-19-2121-workflow-engine/spec.md:230](docs/dev-sessions/2026-05-19-2121-workflow-engine/spec.md) says workflow engine tools register with \`priority: critical\` (always-loaded). The PR currently ships them with \`priority: normal\` to satisfy \`TestCoreToolsDeclarePriority\` after an intentional edit removed hardcoded \`critical\` from \`src/decafclaw/tools/workflow_tools.py\`.

**Implication:** under token-budget pressure, the static workflow tools (\`workflow_start/list/switch/status\`, \`workflow_artifact_read/write\`) can be deferred from the catalog. \`workflow_status\` in particular is the recovery tool after compaction; losing it just when catalogs get noisy is the bad case.

The dynamic \`phase_advance\` is unaffected — it's injected via \`ctx.tools.extra_definitions\` by \`refresh_workflow_tools\`, bypassing the priority classifier.

**Decide:**
- Set all 6 \`WORKFLOW_TOOL_DEFINITIONS\` to \`priority: critical\` per spec, OR
- Keep \`normal\` and update the spec text to match, OR
- Use \`config.agent.critical_tools\` allowlist as the configurable mechanism instead

**Files:**
- \`src/decafclaw/tools/workflow_tools.py:312, 330, 347, 363, 374, 392\` — static defs
- \`src/decafclaw/tools/workflow_tools.py:77\` — dynamic \`phase_advance\` definition (also missing priority)

lmorchard:
## Summary

Implements [#255](https://github.com/lmorchard/decafclaw/issues/255) — declarative multi-phase workflows scoped to a single conversation. Workflows are `kind: workflow` skills with `phases/*.md` files declaring tools, transitions (with `when:` annotations the LLM uses to route), edge-level review gates, inline + subagent phase kinds, and per-phase context-composer overrides.

Originally designed with cross-conversation runs (run-ids, `workflow_list`/`workflow_switch`). Demos surfaced cascading wiring bugs from the layered state machine; PR was reworked in place to scope state to one conversation, simplifying the engine, eliminating the run-id namespace, and resolving the demo failure mode. See [`docs/dev-sessions/2026-05-21-1732-workflow-engine-rework/spec.md`](docs/dev-sessions/2026-05-21-1732-workflow-engine-rework/spec.md) for the rework rationale and [`plan.md`](docs/dev-sessions/2026-05-21-1732-workflow-engine-rework/plan.md) for the 9-task execution.

### What ships

- `src/decafclaw/workflow/` module (types, loader, **conv_state**, engine, subagent, context overlay, registry)
- `src/decafclaw/tools/workflow_tools.py`: `workflow_start` / `workflow_status` / `workflow_abort` + `workflow_artifact_{read,write}` + dynamic `phase_advance` (with `priority: critical`)
- Skill loader recognizes `kind: workflow` and registers loaded definitions
- `required-skills:` in SKILL.md frontmatter; engine auto-activates them on `workflow_start` (via the standard skill-tier approval flow)
- `ContextComposer` consults a `WorkflowOverlay` in INTERACTIVE mode and applies per-phase context-profile overrides; per-iteration `workflow_msg` injection (workflows started mid-turn are visible in the next iteration)
- Per-phase tool catalog hard-gate via `ctx.tools.allowed`
- Subagent dispatch via `ConversationManager.enqueue_turn(kind=CHILD_AGENT)`; child inherits parent's skill tools AND parent's `conv_id` so artifact writes resolve to the parent's directory
- Conv-scoped persistence at `conversations/{conv_id}/workflow.json` + `artifacts/`; one active workflow per conversation, sequential after `workflow_abort` or terminal state
- Bundled `research_brief` demo workflow with `required-skills: [tabstack]` exercising every engine feature
- 79 workflow-specific tests; full suite 2777 passing
- `docs/workflows.md` documenting authoring + run state + limitations

### Resolved follow-ups

- **#561** (`phase_advance` critical priority) — resolved here. The dynamic schema now declares `priority: critical` so the tool stays in the active catalog under deferral pressure. **Closes #561.**

### Carried-forward follow-ups

- **#562** (engine writes phase-boundary markers) — composer reads them; engine writes none yet (no-op until wired).
- **#563** (`decision-slice: off` context-profile override) — cross-subsystem with compaction.py.
- **#564** (subagent-skill integration test) — escape hatch supported in loader + subagent, no real-skill round-trip test.

## Test plan

- [x] `make check` — clean (ruff + pyright + tsc)
- [x] `make test` — full suite 2777 passing
- [ ] Manual smoke: `/research_brief start <topic>` walks gather → draft → review → publish
- [ ] Eval at `evals/workflow_routing.yaml`


lmorchard:
# Workflow engine (#255): retrospective + a code-driven redesign

*Side analysis (not a directive to whoever's actively working this PR — a
second opinion drawn from a separate code-driven research agent, "Sophie").
Summarizes the three iterations here, diagnoses why it hasn't run end-to-end
against a live LLM yet, and proposes a salvage-oriented reshape of the
execution kernel. Take or leave per your own context on the current work.*

---

## TL;DR

Three iterations and ~12.7k lines into PR #557, the engine has **never walked
a workflow end-to-end against a real LLM.** Every iteration passed every test
that didn't need an LLM and stalled the moment one did. The root cause isn't a
wiring bug — it's the architecture: **the LLM is the thing that drives the
state machine forward.** Forward progress depends on the model choosing to emit
a specific tool call (`phase_advance`, `workflow_artifact_write`), and it
reliably doesn't.

We already have two existence proofs sitting in our own tree of how to do this
right, and they agree:

- The **`project` skill** works because **the human drives** phase transitions.
- **Sophie** (my research agent) works because **code drives** phase transitions.

The thing that doesn't work — the workflow engine — is the one where the **LLM
drives** transitions. The fix is to stop asking the LLM to drive the machine.
Make the engine own control flow and invoke the LLM as a structured-output
*worker* for each step. This is exactly what #255's original references already
told us ("Building Effective Agents": workflow vs. agent; CoALA:
environment-controlled loops). The vision was right; the implementation drifted
into an agent loop wearing a workflow costume.

The fix is more surgical than it first looks. PR #557 is **mostly salvageable**:
the declarative format, loader, and even the engine's transition mechanics are
sound — and the subagent dispatch→verify→advance path is *already* code-driven
and is the template for the redesign. The one thing to remove is the LLM-emitted
`phase_advance`-as-crank for inline phases (and the not-yet-built nudge loop
proposed to patch it). Replace it by generalizing the pattern subagent phases
already use.

---

## What happened across three iterations

1. **Cross-conversation engine** (`2026-05-19-2121-workflow-engine`) — run-ids,
   `workspace/workflows/{name}/runs/{run-id}/`, `workflow_list`/`workflow_switch`.
   Live smoke exposed cascading wiring bugs from the layered state machine.

2. **Conversation-scoped rework** (`2026-05-21-1732-workflow-engine-rework`) —
   dropped run-ids and cross-conv discovery, moved state to
   `conversations/{conv_id}/workflow.json`, added `workflow_abort`,
   `required-skills`, `phase_advance` priority:critical. Engine plumbing went
   clean; 79 workflow tests green. **Then the first live LLM smoke stalled in
   iteration 2** — the LLM called `workflow_start`, narrated "Okay, I've started
   the workflow," and ended the turn. The machine was correct; nobody drove it.

3. **Phase-turn model** (`2026-05-29-1729...`) + **cheap-experiment**
   (`2026-05-29-1803...`) — correctly diagnosed "the LLM doesn't drive the
   workflow forward" and proposed flipping to "engine drives flow, LLM drives
   routing." Before committing to that reach, we ran a prompt-only experiment
   (make the tool result loud). Four real-LLM smokes, three distinct stall
   modes, all the same shape:

   > "Gemini Flash treats 'produce output X' as 'write text in your response.'
   > Tool calls are a secondary affordance the model has to be bullied into
   > using."

   The cheap experiment's own conclusion: prompt-only is insufficient; the
   bottleneck is structural.

The honest summary: we have spent three cycles **generalizing the authoring
surface** (declarative phase graphs, edge gates, `when:` routing, dynamic tool
schemas, context overlays, subagent dispatch, persistence) while the
**execution kernel has never once run** for even a single hardcoded workflow.
That's building the plugin API before the engine turns over.

---

## The actual root cause: who drives the transitions?

Every multi-phase system has to answer one question: **what causes the machine
to move from phase N to phase N+1?** There are exactly three possible drivers.

| Driver | Example in our tree | Works? | Why |
|---|---|---|---|
| **Human** | `project` skill (brainstorm→spec→plan→execute) | ✅ | The user paces transitions: "looks good, make a plan." The agent does work *within* a phase; the human moves between phases. |
| **Code** | Sophie research orchestrator | ✅ | A plain `while` loop sequences phases. The LLM is called only for bounded structured-output tasks; code reads the results and decides the next step. |
| **LLM** | Workflow engine (PR #557) | ❌ | The machine only advances when the model emits `phase_advance`. Forward progress is contingent on the least reliable component choosing to fire a specific tool. |

This reframes our entire stall taxonomy. "Narrate instead of call,"
"inline-content instead of `artifact_write`," "acknowledge and stop" are not
four bugs — they're one structural fact: **we made the LLM responsible for
turning the crank, and it has no reliable reason to keep turning it.**

The phase-turn spec is reaching toward the right answer ("engine drives flow")
but stops half-way: it keeps `phase_advance` as a tool call the LLM must emit,
then bolts on a `max_phase_continuations` nudge-loop to retry when it doesn't.
That's "LLM in the loop, patched" — not "LLM out of the loop." We'd be shipping
a retry loop around the exact unreliability that's been failing every smoke.

---

## What Sophie does (the code-driven model in detail)

Sophie is a multi-phase research agent (plan → iterate[search→fetch→analyze→
evaluate] → outline → write → judge). It runs reliably. The reasons map
directly onto our problem:

1. **The LLM is never in the control-flow loop.** The orchestrator is plain
   imperative code — an explicit `while iteration <= maxIterations` with
   explicit phase ordering and explicit stop conditions. The model is invoked
   only for discrete, single-shot tasks.

2. **Every LLM call is structured output, not a tool the model may or may not
   fire.** Planning, analysis, gap-evaluation, outline, judgment all use
   schema-constrained structured generation. The model *cannot* "narrate
   instead" — the call's contract **is** the structured shape. There's no
   "did it remember to call the tool?" failure mode because emitting the result
   is the only thing the call can do.

3. **The one place the model influences flow, it advises — code decides.** Gap
   evaluation returns `shouldContinueResearch: boolean` plus next queries; the
   *orchestrator* reads that field and decides whether to loop. And it's
   defensively distrustful: if the model says "continue" but supplies no
   queries, the code **overrides it to stop.** The model is a sensor, not the
   driver.

4. **Phase output is a return value the engine captures — not a side effect the
   model must remember to perform.** Each phase returns its product (analyzed
   page, outline object); the orchestrator threads it forward. Compare smoke #4,
   where our gather subagent did the research and then *wrote the markdown into
   its text response* instead of calling `workflow_artifact_write`. Sophie
   would have captured that text **as** the artifact.

5. **Validate-and-fallback, not nudge-and-pray.** Failure-rate thresholds,
   citation-count retries, hallucinated-ID filtering, query clamping, graceful
   degradation. When output is missing or malformed, code takes a defined path
   rather than re-prompting and hoping for a better draw.

---

## Proposed redesign

### Principle

> **The engine owns control flow and invokes the LLM as a structured-output
> worker for each step. The LLM is never the component that has to remember to
> advance the machine.**

This keeps everything you actually want from the phase-turn model — phases live
in the conversation timeline, users can interrupt, turns are observable — while
removing the failure mode all four smokes hit.

### Five concrete changes

**1. Auto-advance every deterministic edge in code.** Look at the demo:
`gather → draft → review → publish`. Of four phases, exactly **one** (`draft`)
has a genuine multi-edge decision (`review` vs. back-to-`gather`). The rest are
single-edge or terminal. We *already* auto-advance single-edge **subagent**
phases (`engine.dispatch_and_finalize_subagent`). Generalize that to **all**
single-edge phases and ~75% of the demo graph stops depending on the LLM
emitting anything.

**2. Make routing a structured-output call at real branches — not a tool call
we nudge.** Where a phase genuinely forks, after the work turn the engine makes
a *separate, schema-constrained* decision call:

```python
decision = await llm.structured(
    schema={"target": Enum[<phase.next_phase_ids>], "reason": str},
    context=phase_work_so_far,
    instructions=render_when_clauses(phase.next_phases),  # your `when:` text
)
engine.advance(state, decision.target, decision.reason)
```

The model physically cannot stall here — structured output is the only thing
the call produces. Our existing `EdgeDef.when` annotations are *already* the
perfect enum-option descriptions for this call. This is Sophie's gap-evaluation
pattern, and it's the reliable version of what the nudge-loop approximates.

**3. Capture phase work product as a return value.** A phase that "produces
sources.md" should have its turn output captured **as** that artifact by the
engine, rather than depending on the model routing content through a write
tool. The `outputs:` declaration + filesystem-verify dance is compensating for
a contract we don't actually control — invert it so the contract is the engine's.

**4. Keep humans and gates as first-class drivers.** This is why the `project`
skill works. A gate (review) is a human-driven transition; an interview phase
is a human-paced one. The engine should treat "wait for the human" as a normal,
expected driver — not a stall to be nudged out of. Phases that need user input
simply don't auto-advance; the next user message drives them. (The phase-turn
spec already gets this part right.)

**5. Validate-and-fallback at every boundary.** Missing artifact, malformed
routing decision, subagent crash → defined fallback (re-run once with a tighter
instruction, route to an error phase, or surface to the user), not an open-ended
nudge loop. Bound everything.

### Build order (the part I feel strongest about)

**Prove the kernel before generalizing it.** Write `research_brief` as a plain
imperative async function — `gather()`, `draft()`, `review()`, `publish()` —
where the engine calls the LLM per step with structured output, code captures
each result, code sequences the phases, and the *only* LLM decision is `draft`'s
one branch (via a structured routing call). No declarative loader, no graph, no
dynamic tool schemas in the path. Get **that** walking reliably against Gemini
Flash, in the conversation timeline, interruptible.

*Then* lift the declarative format (`kind: workflow`, `phases/*.md`,
`next-phases`, `when:`, gates) back on top of a proven execution model. We have
the authoring format and loader already — they're fine and worth keeping. What
we don't have is a kernel that runs, and we've been decorating its exterior for
three iterations.

---

## Salvage triage: what's keep / change / drop in PR #557

This is **not** a rebuild. Most of #557 survives. The change is surgical because
the engine *already contains* the code-driven pattern we want — subagent phases
dispatch→verify→advance with no LLM `phase_advance` call. We're generalizing
that, not replacing the engine.

**Keep as-is (sound, even exemplary):**
- The authoring format: `kind: workflow` skills, `phases/*.md`, frontmatter
  (`tools`, `next-phases`, `when:`, `gate`, `outputs`, `context-profile`).
- `loader.py`, `types.py`, `registry.py` — the declarative → dataclass layer.
- `conv_state.py` — conv-scoped persistence + `conv_lock`.
- **Engine transition mechanics**: `advance()`, `_apply_transition`,
  `_enter_gate` / `finalize_gate_response`. Pure, correct primitives. What
  changes is *who calls them*, not the functions themselves.
- **The subagent `dispatch_and_finalize_subagent` → `verify_subagent_outputs`
  → auto-advance pattern.** This is the model for the redesign — keep it and
  extend it to all phases.
- `required-skills:` auto-activation, per-phase tool gating via
  `ctx.tools.allowed`, edge gates via `EndTurnConfirm`.
- The phase-turn model's turn-scheduling instinct (engine enqueues turns,
  phases live in the timeline, users interrupt), `parent_conv_id`,
  `WORKFLOW_PHASE` compose mode — the right surface, compatible with this.

**Change (the narrow fix):**
- Inline phases stop advancing via LLM-emitted `phase_advance`. Engine
  auto-advances single-edge phases (the subagent pattern, generalized); a
  *structured routing call* decides multi-edge branches. `advance()` is
  untouched — only its caller flips from "tool the LLM fires" to "engine code."
- Phase output moves from "model remembers `workflow_artifact_write`" to
  "engine captures the turn's result as the artifact." Helps subagents too.

**Drop (don't build it):**
- The `max_phase_continuations` nudge loop from the phase-turn spec. It's a
  patch around the exact crank we're removing. Not yet implemented — leave it
  unimplemented.

## What this is NOT (honest caveats)

- **Not "make decafclaw a batch pipeline."** Sophie has no human in the loop, no
  conversational surface, no interruption. We want all three. The borrowed
  lesson is narrow: *where control flow lives and how the LLM is invoked* — not
  "remove the chat."
- **Not "the LLM never decides anything."** It still makes routing decisions at
  genuine forks — but as structured output, not as a hoped-for tool call.
- **Not free.** A structured routing call per branch is an extra LLM round-trip.
  But it's a *reliable* round-trip, versus a nudge loop that's an unbounded
  number of unreliable ones.

## Recommendation

1. **Don't abandon #557 — rework it surgically.** Keep loader/types/registry/
   conv_state and the engine's transition mechanics. Retire only the
   LLM-driven `phase_advance`-as-crank for inline phases, and don't build the
   nudge loop.
2. **De-risk with a throwaway spike first.** Prove the code-driven pattern on a
   hardcoded `research_brief` — engine sequences phases, structured routing at
   the one branch, captured outputs — directly in the conversation timeline,
   reliable end-to-end against Flash (the bar no iteration has cleared). This
   can be a standalone script; its job is to validate the mechanism cheaply.
3. **Fold the proven pattern into the existing engine.** Generalize the subagent
   dispatch→verify→advance path to all phases; add output capture; replace
   `tool_phase_advance`'s caller role with engine-driven advance + structured
   routing. The declarative layer rides on top unchanged.

If the spike clears the bar, we'll have proven the mechanism in ~150 throwaway
lines — then applied it to the engine we already have, generalizing from
something that works instead of something that doesn't.

lmorchard:
Pulling back from the phase-turn rework based on this analysis. The diagnosis lands: every iteration kept the LLM responsible for emitting `phase_advance`, every smoke stalled on that exact crank, and the phase-turn spec's nudge-loop would have wrapped retries around the failure mode rather than removed it.

Pivot captured at [`docs/dev-sessions/2026-05-31-1223-code-driven-engine-spike/notes.md`](docs/dev-sessions/2026-05-31-1223-code-driven-engine-spike/notes.md) (commit `024605e`). Phase-turn `spec.md` + `plan.md` get redirect banners; cheap-experiment commits stay on the branch as history. Task ahead is the ~150-line imperative spike of `research_brief` — engine sequences phases in code, structured-output LLM calls per step, only the `draft` branch is an LLM routing decision (via schema, not tool emit). Fresh session for that work.

Thanks for the second look — it caught us before we built another reroll of the same problem.

lmorchard:
Closing in favor of a fresh approach to #255.

After four iterations here (cross-conv → conv-scoped → phase-turn → code-driven spike), the lesson is settled: a workflow needs a **programmatic control scaffold**. Code must own control flow; the LLM is invoked only as a constrained structured-output worker. This engine's core — LLM-driven phase transitions (`phase_advance` as the crank) — is the thing that never reliably walked end-to-end against a live model, and the spike confirmed the code-driven alternative works first try.

The new direction (spec: `docs/dev-sessions/2026-06-05-1455-workflow-replay-engine/spec.md` on branch `feat/255-workflow-replay`) is **durable execution via deterministic replay**: a workflow is a first-class `TurnKind.WORKFLOW` orchestrator written as plain async Python, with two journaled primitives (`llm_call`, `user_input`). The engine re-runs the function from the top on resume, replaying journaled results — which gives durable, restart-surviving, human-in-the-loop workflows where control flow is just Python (`if`/`while`), not a DSL.

The engine here is replay-shaped vs. graph/DSL-shaped — different enough that a fresh branch is cleaner than reworking ~15k lines. These commits stay as the record of what didn't work; ideas (conv-scoped persistence + lock, `RunStatus` lifecycle) carry forward as reference. The Sophie analysis that drove this is in `docs/workflow-engine-255-redesign-proposal.md`.

PR #578, opened 2026-06-11
feat(conversations): migrate sidecars to conversations/{conv_id}/ directory layout

### LINKED ISSUE #576: Conversation sidecars: migrate to `conversations/{conv_id}/` directory layout
## Context

Per-conversation sidecar files live in `workspace/conversations/` in two conventions today:

**Flat (legacy):**
- `{conv_id}.jsonl` — archive
- `{conv_id}.notes.md` — per-conv notes
- `{conv_id}.decisions.json` — compaction decision slice
- `{conv_id}.context.json` — context-composer diagnostics

**Directory (newer, since PR #573):**
- `{conv_id}/workflow.json` — workflow replay journal

Two conventions for the same logical grouping — "everything that belongs to this conversation" — is friction we should resolve.

## Why migrate

- **Listing.** `ls conversations/` returns mixed flat files + dirs. A reader has to know which is which.
- **Future sidecars.** Whatever ships next (`#574` adds journals for batch primitives — could be one file or many) deserves a single answer, not "depends when the feature shipped."
- **Cleanup story.** Deleting a conversation today means globbing for sidecars. Directory layout makes it one `rm -rf`.

## What this touches

Every reader/writer of a per-conv sidecar:
- `archive.py` — `read_archive`, `append_message`, archive_path helpers
- `notes.py` — notes_path
- `compaction_decisions.py` — decisions sidecar
- `context_composer.py` (diagnostics sidecar)
- `startup_scan` archive glob in `conversation_manager.py` (currently `conversations/*.jsonl`)
- Plus any tests that hand-construct paths

## Backwards compatibility

Existing user data has flat sidecars for every prior conversation. Options:

1. **Lazy migration on next access.** Readers try the new path, fall back to the old, migrate on next write. Slow to converge but no downtime / no big bang.
2. **One-shot migration script.** Run once at startup or via `make migrate-sidecars`. Move all files, done. Cleaner but requires the script to be correct on first try.
3. **Hybrid.** One-shot for existing data, lazy for safety. Code carries the legacy reader for ~one release cycle then it's removed.

(3) is probably right — matches how decafclaw has handled other data-shape changes.

## Acceptance criteria

- Every sidecar reader/writer goes through a single helper that returns `conversations/{conv_id}/{file}` paths.
- A migration utility moves existing flat sidecars into the directory layout. Idempotent — safe to re-run.
- `startup_scan` walks the new layout (and falls back to flat during the deprecation window).
- Tests use the directory layout; no remaining hand-constructed `.jsonl` paths in test code.
- Docs: `docs/data-layout.md` updated; CLAUDE.md key-files note (if any path mentions get specific) reviewed.

## Related

- #255 (workflow design)
- #573 (replay engine PR — introduced the directory convention for `workflow.json`)

lmorchard:
## Summary
- Moves every per-conversation sidecar from the flat `conversations/{conv_id}.SUFFIX` layout into a per-conversation directory `conversations/{conv_id}/{file}`.
- Why: one listing shows all of a conversation's state, one `rm -rf` deletes it, and future sidecars get one answer instead of "depends when the feature shipped." Resolves the two-conventions friction introduced when #573 added `workflow.json` under a directory.

## Design Decisions
- **All 9 flat sidecars migrate, not the issue's 4.** Research found the codebase has 9 flat sidecar types (archive, compacted, notes, decisions, context, canvas, skills, skill_data, vault_grants), 4 archive-discovery sites (not 2), and a delete site hardcoding only 5 of 9 suffixes. Migrating a subset would leave the exact mixed layout the issue exists to remove.
- **Single chokepoint.** New `conversation_paths.py` owns path construction + sanitization: `conversation_dir`, `sidecar_path`, `iter_conversation_archives`, `delete_conversation_files`, and `SIDECAR_FILENAMES`/`SIDECAR_LEGACY_SUFFIXES`. `workflow/paths.py` delegates to it (the duplicated `_safe_conv_id` is gone).
- **In-place fallback for reads AND writes** (not "writes always new"). While a flat legacy file exists and the new-layout file doesn't, code keeps reading *and writing* the flat file, so an existing archive never splits across two locations. No code path moves files — only the migration script does.
- **Hybrid migration.** `scripts/migrate_sidecars_to_dirs.py` (idempotent, `--dry-run`) + `make migrate-sidecars` / `make migrate-sidecars-dry`. The flat read-fallback stays for one deprecation cycle (follow-up issue to remove it).

## Changes
- **New:** `src/decafclaw/conversation_paths.py`, `scripts/migrate_sidecars_to_dirs.py`, Makefile targets.
- **Routed through the helper:** archive/compacted (`archive.py`), notes, decisions (`compaction_decisions.py`), context diagnostics (`context_composer.py`), canvas, skills/skill_data (`persistence.py`), vault_grants.
- **Discovery sites** now use `iter_conversation_archives`: `startup_scan`, `conversation_search`, `list_system_conversations`, newsletter scheduled-activity.
- **Delete** (`http_server.py`) uses `delete_conversation_files` — removes the whole `{id}/` dir + all 9 flat legacy suffixes, fixing a pre-existing leak (canvas/skills/skill_data/vault_grants were orphaned on delete).
- **Docs:** `data-layout.md`, `context-composer.md`, `CLAUDE.md` (+ key-files entry), and ~9 doc/docstring path references.

## Test Plan
- [x] `make test` — 2859 passed (+39: new `conversation_paths`, migration, search-tool tests; no-split-archive invariant; per-module fallback)
- [x] `make check` — lint + pyright + tsc clean
- [x] Migration script proven on throwaway dirs incl. the `.compacted.jsonl`-before-`.jsonl` ordering trap and idempotency
- [ ] Manual: run `make migrate-sidecars-dry` against a copy of real data to sanity-check the mapping, then `make migrate-sidecars` (after a backup) — **owner's call; not run here** since the worktree shares the live `data/` dir

## References
- Spec: `docs/dev-sessions/2026-06-10-1728-conversation-sidecar-dirs/spec.md`
- Plan: `docs/dev-sessions/2026-06-10-1728-conversation-sidecar-dirs/plan.md`
- Notes (incl. deferred follow-ups): `docs/dev-sessions/2026-06-10-1728-conversation-sidecar-dirs/notes.md`
- Related: #255, #573

Closes #576


lmorchard:
Addressed both Copilot comments in `conversation_paths.py` (fail-open is the project convention for these read paths):

- `iter_conversation_archives`: `base.iterdir()` now wrapped in `try/except OSError` → logs a warning and yields nothing, so a transient FS error can't break startup recovery / search / UI listing. Added `test_iter_archives_fails_open_on_oserror`.
- `delete_conversation_files`: the legacy-sidecar `unlink()` loop now catches `OSError` + logs, matching the `rmtree` handling.

`make test` 2860 passed, `make check` clean. Force-pushed.

Issue #575, opened 2026-06-11
Workflow turns: reconsider Context kind (USER vs `Context.for_task`)


lmorchard:
## Context

PR #573 shipped `TurnKind.WORKFLOW` reusing the USER `Context` path — full interactive Context with per-conversation state restored. The dispatch site in `conversation_manager.py:1396-1411` explicitly notes this as a "revisit after live smoke" decision:

> WORKFLOW deliberately reuses this path for now: the orchestrator only needs `ctx.config` + `ctx.conv_id`, and running on the user conv keeps phases visible in the timeline. (Revisit after live smoke: WORKFLOW may want `Context.for_task` semantics — if so, add it to `TASK_KINDS` with a `KIND_TASK_MODE` entry.)

Live smoke is now PASS. Time to make the call.

## The question

Should WORKFLOW turns join `TASK_KINDS` (alongside `WAKE`, `SCHEDULED`, `CHILD_AGENT`) and use `Context.for_task` semantics, or stay on the USER `Context` path?

## Trade-offs

**`Context.for_task` would give workflows:**
- `skip_reflection=True` by default — workflows are code-driven; reflection isn't applicable
- `skip_vault_retrieval=True` by default — workflows don't auto-browse the vault
- Cleaner per-turn state (no per-conv USER state carried in)
- Consistency with other harness-orchestrated turn kinds

**USER context (status quo) preserves:**
- Per-conversation skill activation state (irrelevant for the engine, but might matter if a workflow ever wants to inspect what skills were live)
- Active-model selection from the conversation (a workflow's `wf.llm_call(model=…)` is currently independent of this — see the per-call model override note in `handle.py`)
- The conversation's vault-retrieval mode (currently no effect on workflow turns)

## What the live smoke surfaced

Nothing — the interview workflow only touches `ctx.config` and `ctx.conv_id`. The question is forward-looking:

- Once #574 (batch primitives — `subagent` / `parallel` / `pipeline` / `tool_call`) lands, the `subagent` primitive will need to fork a child Context. The parent's Context kind shapes the child's defaults.
- Once a second hero workflow ships with more substantive LLM use, reflection-on-workflow-turns could surface as either spurious noise (argues for task-mode) or useful (argues for USER).

## Acceptance criteria

A small ADR-style decision committed alongside whatever code change (or non-change) results. Either:

[...]

Issue #564, opened 2026-05-20
Workflow: integration test for subagent-skill: escape hatch


lmorchard:
Follow-up from #557 (workflow engine PR).

The workflow engine's \`subagent-skill: <name>\` escape hatch lets a subagent phase activate a named skill instead of using an inline prompt body. Implementation:
- \`src/decafclaw/workflow/loader.py\` parses the field
- \`src/decafclaw/workflow/subagent.py:_run_child\` looks up the skill, uses its body as the system prompt, and raises if the body is empty (defends against lazy-loaded skills not yet activated)

**Gap:** no integration test exercises this path with a real bundled skill. Current tests cover the loader (\`tests/test_workflow_loader.py:test_load_subagent_skill_escape_hatch\` — verifies parsing) but \`_run_child\` is monkeypatched in engine tests, so the actual lookup-and-use path is untested.

**What to do:**
- Add a test that registers a small bundled skill, creates a workflow with a subagent phase using \`subagent-skill: <name>\`, runs \`dispatch_and_finalize_subagent\` against a mocked \`enqueue_turn\` that captures the child's setup callback, and asserts:
  - The setup callback sets \`child_ctx.config.system_prompt\` to the skill's body
  - The empty-body case raises \`ValueError\`
- Alternatively (or in addition): a real-LLM eval case using a workflow with \`subagent-skill:\` that ships in \`evals/\`

**Files:**
- \`src/decafclaw/workflow/subagent.py:97-110\` — the branch under test

lmorchard:
Superseded — this tracked the LLM phase-state-machine (PR #557), which was *replaced* by the code-driven replay engine (commit 14b24ec, 'step-primitive engine replaces phase engine'). The referenced mechanism (WORKFLOW_TOOL_DEFINITIONS / workflow_phase_boundary / context-profile overlay / subagent-skill loader) no longer exists in the codebase. Closing as obsolete; see docs/workflows.md.

Issue #474, opened 2026-05-13
Mixture-of-agents tool: parallel reference models + synthesizer


lmorchard:
## Context

Tool that calls N reference models in parallel and feeds responses through a synthesizer model. Useful for "give me three takes on X" or consensus on hard calls. Hermes-agent's `mixture_of_agents_tool.py` is a small (~400 LOC) tool that does this via OpenRouter.

## Proposed approach

- New `src/decafclaw/tools/moa.py`.
- `mixture_of_agents(prompt, reference_models=[...], synthesizer_model=...)`.
- Use the existing LLM provider abstraction so it works against Anthropic / OpenAI / OpenRouter — not OpenRouter-only as hermes is.
- Returns the synthesizer's response, with each reference model's raw response available as `ToolResult.data` for inspection.

## Validation

- Test: mock 3 providers, assert all called in parallel, synthesizer sees all 3.
- Smoke: contentious question to test whether MoA gives a more nuanced take than a single call.

## Open questions

- Hardest part is the *description* — getting the LLM to know when to reach for this tool over a regular call. Description-engineering matters more than implementation.
- Niche use case — low certainty this earns its keep. Worth one experiment then keep-or-cut.

## Reference

`hermes-agent/tools/mixture_of_agents_tool.py`.

Issue #421, opened 2026-04-29
Brainstorm: shared DAG substrate across workflows, notebooks, dashboards (sister to #255)


lmorchard:
## Status

**Brainstorm — expected to iterate.** Sister issue to #255 (workflows). This isn't a commitment to build anything; it's a place to capture the observation that several speculative features may want the same underlying primitive, so we don't accidentally build three incompatible engines. Add comments / edit the body as research and ideas accumulate.

## The observation

Three open ideas appear to want the same underlying thing:

1. **Workflows** (#255) — a graph of phases with allowed tools per phase, gated transitions, and persistent state. LangGraph-style.
2. **Notebooks** (#420, Path A & B) — a 1-D sequence of cells where each cell's output feeds the next, with cascading re-evaluation when an upstream cell changes.
3. **Dashboards** (#420 sister concept) — a 2-D arrangement of blocks where each block's data may depend on another's, refreshed on schedule or on demand.

Plus a fourth, already-shipped: **scheduled tasks** — cron-driven invocations of skills/tools, currently flat (no inter-task dependencies).

What they share: **a DAG of typed nodes with gated transitions, persistent state, and observable runs.** The differences are presentation (linear / 2D / state-machine) and trigger (edit / schedule / manual / phase-boundary).

## The question this issue exists to answer

**Does a shared DAG substrate make sense across all four, or are they genuinely different things that just look similar?**

Three possible answers, each with implications:

- **Yes, unified.** Build one DAG runtime (executor + dependency tracking + state + observability) and present it three ways. Maximum leverage; biggest architectural commitment; risk of premature abstraction.
- **Partially unified.** Notebooks and dashboards share a runtime; workflows are a peer concept that *uses* the runtime as one of its node types. Workflows stay state-machine-shaped because phase-transition semantics are different from data-flow semantics.
- **No, three different engines.** Each has different enough semantics (DAG re-execution rules, persistence shape, error handling) that the abstraction would be all-cost no-benefit.

I lean toward **partially unified** based on a hand-wave, but this is exactly the question that should be answered with research and prototyping, not architecture-astronaut design.

## Research action items

Before sketching any concrete design, get hands-on with the existing ecosystem:

- **n8n** — node-based workflow automation. Closest visual analogue to what a "workflow editor" might look like.
- **Node-RED** — flow-based programming for IoT / event-driven workflows. Mature DAG runtime + visual editor.
[...]

PR #579, opened 2026-06-12
feat(workflow): batch primitives — tool_call, subagent, parallel, pipeline (#574)

### LINKED ISSUE #574: Workflow batch primitives — `subagent`/`parallel`/`pipeline`/`tool_call` (the fan-out shape)
## Context

PR #573 (closes #255) shipped the workflow replay engine with **two** journaled primitives: `wf.user_input` and `wf.llm_call`. Those are sufficient for the hero workflow (`/interview`) — a single orchestrator, sequential, one LLM at a time, one user answer at a time.

What they cannot express is **fan-out**: spawning N workers in parallel, pipelining a list of items through a sequence of stages, or calling a tool/subagent as a journaled boundary crossing. The replay model handles these fine in principle — they're just real journaled wrappers — but they're not built yet.

This issue tracks the deferred work as a single batch so it doesn't sit buried in [PR #573's notes](https://github.com/lmorchard/decafclaw/blob/main/docs/dev-sessions/2026-06-05-1455-workflow-replay-engine/notes.md).

## The four primitives

Each is a journaled wrapper that crosses to the outside world — recorded on first execution, replayed from the journal on re-run. Plain Python orchestrates them; the engine does not need a new abstraction.

1. **`wf.tool_call(name, **args)`** — call a registered decafclaw tool from inside an orchestrator. The natural way to make workflows actually *do* things beyond LLM and user input. Currently you can simulate it inside an `llm_call` schema, but a direct primitive is cleaner and avoids smuggling tool behavior through prompt engineering.

2. **`wf.subagent(prompt, *, schema=None, tools=None, ...)`** — dispatch a constrained child agent loop as a single journaled boundary crossing. Returns its final result. Today a workflow's only LLM access is single-shot structured output via `llm_call`; this opens the door to tasks that genuinely need an agent loop (multi-turn tool use, exploration).

3. **`wf.parallel([thunk1, thunk2, …])`** — run N thunks concurrently, await all, return the result list. The barrier shape. Each thunk's journaled calls record under its own positional sub-keys so the journal is still deterministic on replay.

4. **`wf.pipeline(items, stage1, stage2, …)`** — pipeline N items through M stages with no barrier between stages. Item A can be in stage 3 while item B is still in stage 1. Wall-clock = slowest single-item chain, not sum-of-slowest-per-stage. This is the *default* fan-out shape for most multi-stage work; `parallel` is the right call only when a stage genuinely needs cross-item context from all of the prior stage.

## What this unlocks

- **Research / audit / sweep workflows** — fan out N parallel finders, dedupe, fan out N verifiers, synthesize.
- **Multi-step transformations over collections** — migrate-style pipelines where each item flows through transform → verify → write.
- **Quality patterns** that need multiple votes — adversarial verify (N skeptics per finding), perspective-diverse verify (correctness/security/perf lenses), judge panels.
- **Mixed agentic + structured** — an orchestrator can route some sub-problems to a subagent loop (open-ended) and others to a single forced-tool `llm_call` (focused), with the journal making the whole thing durable.

## Design constraints (inherited from the replay engine)

- **Positional keying.** Each journaled call gets a stable position-derived key. Inside `parallel` / `pipeline`, that key has to extend with a sub-index so the journal stays positionally addressable.
- **Determinism guard.** Every entry stores an `args_fingerprint`; a replay mismatch fails loud. Args to `subagent` / `tool_call` / `parallel` / `pipeline` need to fingerprint deterministically — including the *set* of items for `pipeline`, but not their concurrent execution order.
- **No clock / no random.** Same rule as today. If a primitive needs a timestamp, it's journaled.
- **Subagent results must be JSON-serializable.** Same constraint as `llm_call`.
- **Resume crash safety.** Mid-`parallel` crash: the journal records exactly the thunks that completed. Resume re-dispatches the rest. Same as Temporal activity semantics.

## Open questions

- **Cancellation propagation.** If a workflow is cancelled while a `parallel` is running, what's the contract for the in-flight sub-thunks?
- **Subagent context inheritance.** Does the child agent inherit the parent's vault retrieval state, active model, etc., or get a fresh `Context.for_task`?
- **Error policy.** Default for `parallel`: one thunk throws → drop to `null` (Claude Code's choice), or propagate to caller, or configurable?
- **Tool registry visibility.** Should `wf.tool_call` see the same tool registry the agent sees, or a filtered subset configured per-orchestrator?

## Acceptance criteria

- Each primitive has a unit test exercising both the live and replay paths.
- A second hero workflow ships that exercises at least one of the primitives end-to-end (research-brief / audit-sweep / etc. — picked when the design lands).
- Live smoke against `vertex-gemini-flash` walks the new hero workflow with at least one mid-run server restart.
- `docs/workflows.md` "What is not in v1" entry replaced with sections covering each primitive's contract and replay semantics.

## Related

- #255 (workflow design, closed by #573)
- #573 (replay engine PR)
- #421 (shared DAG substrate brainstorm — sister issue)
- #564 (workflow subagent-skill escape hatch test)
- #474 (mixture-of-agents tool — different concept but adjacent)

lmorchard:
## Summary

Extends the workflow replay engine that landed in PR #573 with four new journaled primitives that unlock fan-out workflows: `wf.tool_call`, `wf.subagent`, `wf.parallel`, `wf.pipeline`. Ships a second hero workflow (`/research`) that exercises all four end-to-end, plus a foundation refactor (tuple-path journal keys + sub-handle factory) so primitives compose cleanly.

The fan-out shape was deferred from #573 — see `docs/workflows.md` "What is not in v1" — because it had no concrete use case yet. This PR delivers the use case (research sweeps) and the primitives together.

## Design Decisions

(See `docs/dev-sessions/2026-06-10-1732-574-workflow-batch-primitives/spec.md` for full context.)

- **Hierarchical sub-handles with tuple-path journal keys.** `Journal.entries` becomes `dict[tuple[int, ...], JournalEntry]`; on-disk seqs serialize as dotted strings (`"3.0.0"`). Each `parallel` thunk / `pipeline` item gets a sub-handle keyed `(outer_seq, idx)` with its own cursor — composes naturally with nested fan-out. Existing flat-int journals upgrade transparently.
- **`wf.subagent` wraps `delegate.run_child_turn` directly** (promoted from `_run_child_turn`). Inherits the mature child-dispatch path (restricted tools, no vault writes, child conv_id, timeout). Optional `schema=` for forced structured output, symmetric with `wf.llm_call`.
- **Errors propagate by default** in `wf.parallel`/`wf.pipeline`. Callers wrap individual thunks in try/except for tolerant collection — matches decafclaw's "fail loud" posture.
- **`wf.tool_call` gated by `ctx.tools.allowed`**, mirroring `execute_tool`'s "None means no restriction" semantics. Media stripped from journal entries (large, rebuildable); journal stores only `{text, data}`.
- **Pipeline stage signature is `(prev, item, idx, sub)`** — `sub` is the per-item sub-handle stages MUST use for journaled calls. Departs from Claude Code's Workflow signature deliberately: their workflows aren't durable; ours are, so explicit sub-handle access is required.

## Changes

**`src/decafclaw/workflow/`** — the engine surface:
- `journal.py` (+75): tuple-path keys, dict storage, dotted-string serialization, legacy upgrade.
- `handle.py` (+432): `_next_seq`/`_make_subhandle_at`/`tool_call`/`subagent`/`parallel`/`pipeline`.
- `errors.py` (+23): tuple seq, `WorkflowToolNotAllowed`.
- `resume.py` (+10): tuple seq through confirmation payloads.
- `workflows/research.py` (+167): the second hero workflow.

**`src/decafclaw/tools/delegate.py`** (+129): promoted `_run_child_turn` → `run_child_turn` with unified `(text, data | None)` return; structured-output parsing moved into `run_child_turn`; `ERROR_TEXT_PREFIX` constant.

**Tests** (+50 net, 2820 → 2870): one new test file per primitive plus the orchestrator test, regression test for the parallel exception-unmasking fix, nested key-composition coverage.

**Docs**: `docs/workflows.md` "What is not in v1" replaced with four primitive contract sections + multi-primitive example pointing at `/research`. CLAUDE.md key-files entry updated.

## Test Plan

- [x] `make check` clean (ruff + pyright + tsc).
- [x] `make test` → 2870 passed.
- [x] No new tests in top-25 slowest (`pytest --durations=25`).
- [x] Live walk on `vertex-gemini-flash` with mid-run SIGINT — see `docs/dev-sessions/2026-06-10-1732-574-workflow-batch-primitives/smoke.md` for the full transcript. Journal preserved correctly through the crash; tuple-path keys + sub-handle composition verified on disk.

## Smoke findings (file as follow-up issues)

The live smoke proved the primitives work end-to-end and surfaced three real gaps worth tracking as separate issues (none blocking this PR):

1. **Skill-bundled tools (e.g., `tabstack_research`) aren't reachable from workflow contexts.** Workflow turns don't go through `activate_skill`, so skill tools aren't in the registry; `wf.tool_call` correctly surfaces this as an error-shaped result. Real fix is upstream — pre-activate a configurable skill set when a workflow turn starts, or let `@workflow(...)` declare `requires_skills=[...]`.
2. **No auto-resume on server restart for `status=running` workflows.** Replay machinery is correct (Phase 5/6 unit tests cover full-cache + mid-fan-out resume); the wiring gap is "scan for in-flight journals at startup and re-enqueue."
3. **Unexplained 3-minute hang during pipeline summarize against error-text input.** Could be Vertex throttling on degenerate prompts, could be a primitive issue we missed — needs instrumentation to diagnose.

## References

- Spec: `docs/dev-sessions/2026-06-10-1732-574-workflow-batch-primitives/spec.md`
- Plan: `docs/dev-sessions/2026-06-10-1732-574-workflow-batch-primitives/plan.md`
- Notes (incl. execution-phase decisions and bug-finds): `docs/dev-sessions/2026-06-10-1732-574-workflow-batch-primitives/notes.md`
- Smoke transcript: `docs/dev-sessions/2026-06-10-1732-574-workflow-batch-primitives/smoke.md`
- Related: #421 (shared DAG substrate brainstorm), #564 (workflow subagent-skill escape hatch test), #474 (mixture-of-agents — adjacent concept), #575 (WORKFLOW Context-kind ADR — deliberately deferred).

Closes #574

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -177,7 +177,7 @@ Full doc index: [docs/index.md](docs/index.md). Hot files for navigation:
 `skills/{vault,tabstack,dream,garden,project,claude_code,health,postmortem,ingest,background,mcp,newsletter}/`. `vault`, `background`, `mcp` are always-loaded. Contrib (opt-in) skills live in `contrib/skills/`.
 
 ### Workflows
-`workflow/` — Durable replay engine ([docs/workflows.md](docs/workflows.md)): `registry.py` (`@workflow` decorator), `engine.py` (`run_workflow`), `journal.py` (positional keyed journal + fingerprint), `handle.py` (`WorkflowHandle` — the `wf` object with `llm_call`/`user_input` primitives), `llm.py` (forced-tool structured-output), `errors.py` (`WorkflowSuspended`, `WorkflowNonDeterministic`), `paths.py` (per-conv file paths), `resume.py` (harness glue: `run_workflow_turn` + `WorkflowUserInputHandler`), `workflows/interview.py` (hero orchestrator).
+`workflow/` — Durable replay engine ([docs/workflows.md](docs/workflows.md)): `registry.py` (`@workflow` decorator), `engine.py` (`run_workflow`), `journal.py` (positional keyed journal + fingerprint), `handle.py` (`WorkflowHandle` — the `wf` object with `llm_call`/`user_input`/`tool_call`/`subagent`/`parallel`/`pipeline` primitives), `llm.py` (forced-tool structured-output), `errors.py` (`WorkflowSuspended`, `WorkflowNonDeterministic`, `WorkflowToolNotAllowed`), `paths.py` (per-conv file paths), `resume.py` (harness glue: `run_workflow_turn` + `WorkflowUserInputHandler`), `workflows/interview.py` (suspend/resume hero), `workflows/research.py` (multi-primitive hero — fan-out + pipeline + subagent).
 
 ### Other
 - `prompts/` — System prompt assembly
--- a/docs/dev-sessions/2026-06-10-1732-574-workflow-batch-primitives/notes.md
+++ b/docs/dev-sessions/2026-06-10-1732-574-workflow-batch-primitives/notes.md
@@ -1,0 +1,68 @@
+# Notes — Workflow batch primitives (#574)
+
+## Session start — 2026-06-10 17:32
+
+- Branch: `feat/574-workflow-batch-primitives` (from `origin/main` @ `024175d`)
+- Worktree: `/Users/lorchard/devel/decafclaw/.claude/worktrees/feat-574-workflow-batch-primitives`
+- Session dir: `docs/dev-sessions/2026-06-10-1732-574-workflow-batch-primitives/`
+- HTTP_PORT: 18892 (main uses default 18880, conversation-sidecar-dirs worktree uses 18891)
+- Baseline: `make test` — 2820 passing in 52s
+
+## Pre-flight reading queue (per user)
+
+1. `gh issue view 574` — done; sketch in spec.md
+2. `docs/workflows.md` — engine substrate (research.md §5 captures the "What is not in v1" verbatim)
+3. `docs/dev-sessions/2026-06-05-1455-workflow-replay-engine/` — prior session's spec/plan/notes (research.md §5)
+4. PR #573 — what just shipped + smoke checklist
+
+## Brainstorm — design decisions
+
+### Round 1 (locked)
+
+- **Sub-keying:** hierarchical sub-handles. `wf.parallel` allocates outer seq N, each thunk gets a sub-handle keyed by `(N, thunk_idx)` with its own `_cursor` starting at 0. Journal entry `seq` becomes a tuple-path. Outer entry caches the aggregate result for fast-path full-replay; child entries enable mid-parallel resume. Nests naturally for parallels-of-parallels and pipelines-of-parallels.
+- **Subagent dispatch:** wrap `delegate._run_child_turn` directly (not via tool). Fingerprint includes prompt + schema + allowed_tools. Journal stores text + optional structured data. Reuses the mature child-dispatch path.
+- **Error policy in `wf.parallel`:** propagate (raise out) by default. Callers wrap individual thunks in try/except for tolerant collection. Matches decafclaw's "fail loud" posture.
+- **Tool visibility for `wf.tool_call`:** parent ctx's `allowed_tools` — same set the orchestrator's parent agent could invoke at this turn. Uses existing `tools.execute_tool` plus current allowed_tools gate.
+
+### Round 2 (locked)
+
+- **Cancellation propagation:** cooperative cancel via `ctx.cancelled`. Mirrors `tool_execution.py:307-311`'s pattern — outer cancellation cancels in-flight sub-thunks; completed journal entries are kept; resume continues from where it stopped.
+- **Pipeline stage signature:** `async def stage(prev, item, idx, sub)`. Originally locked as `(prev, item, idx)` in round 2 to match Claude Code's Workflow tool, but plan self-review surfaced that durable replay requires stages to access a per-item sub-handle for journaled calls. Departed from Claude Code's signature deliberately — their workflows aren't journaled, ours are.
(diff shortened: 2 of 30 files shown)


lmorchard:
Addressed all four Copilot inline comments in `35c9b20`:

1. **`Journal.to_dict` sorts by seq.** On-disk entries array is now stable under out-of-order parallel branch completion — no noisy diffs, easier manual inspection. Regression test in `test_to_dict_sorts_entries_by_seq`.

2. **`wf.tool_call` allowlist gate before cursor advance.** Allowlist check moved above `_next_seq()` so a caught `WorkflowToolNotAllowed` doesn't desynchronize the journal on replay. Honors the docstring's "doesn't burn a journal slot" claim. Regression in `test_tool_call_disallowed_does_not_advance_cursor`.

3. **`wf.subagent` defaults model to `self._model`.** Replaced `model or ""` with `model or self._model` so workflows run under their own configured model, matching `wf.llm_call`'s convention. Regression in `test_subagent_defaults_to_handle_model`.

4. **`/research` fails fast on all-error search results.** Added `_is_error_result(r)` helper and an "all results errored → raise" gate after `wf.parallel`. Skill-bundled `tabstack_research` not being reachable from workflow contexts now surfaces as an immediate `RuntimeError` with the underlying error message, rather than wasting tokens on summarization + subagent over `[error: ...]` text. Regression in `test_research_fails_fast_when_search_tool_returns_all_errors`.

Test count: 2870 → 2874 (+4 regression tests). `make check` clean.

Issue #582, opened 2026-06-12
Investigate 3-minute hang in wf.pipeline summarize stage against error-text input


lmorchard:
## Context

During the live smoke for #574 ([PR #579](https://github.com/lmorchard/decafclaw/pull/579), Finding 3), the `/research` workflow appeared to hang for 3+ minutes after `wf.parallel` completed with 5 error-shaped tool_call results (`[error: unknown tool 'tabstack_research']` — see #580). The hang was at the boundary between `wf.parallel` and the first `wf.pipeline` summarize stage; no further journal writes, no Vertex POSTs in the log, server at 0% CPU.

PR #579's fail-fast guard in `/research` masks this for the research workflow specifically, but the underlying hang is unexplained and could bite other workflows that legitimately summarize degenerate input.

## What's known

- All 5 children at `(3, 0..4, 0)` recorded successfully (file mtime `17:02:34` from the smoke log).
- Outer `(3,)` parallel entry NEVER landed.
- No `Vertex generateContent` POST in the server log after the plan stage's 200 OK.
- Server process was alive at 0% CPU for 3+ minutes.

## Hypotheses

- **Vertex throttling on a degenerate prompt.** `_summarize_stage` would have fed `[error: ...]` text to `sub.llm_call`; Vertex may handle this oddly (silent timeout, internal retry). Test by patching `_summarize_stage` with synthetic error-text input and timing the call.
- **`wf.parallel` outer-entry write blocked.** Less likely — `save_journal` is a synchronous atomic write, but the file mtime stayed at the last child's. Could be checked by adding a `log.debug("parallel returning %d results", len(results))` immediately before `journal.append` and re-running.
- **Async deadlock in `_run_concurrent_tasks` cleanup.** The cancel_watcher pattern lives in both `wf.parallel` and `wf.pipeline`; a subtle gather/cancel sequencing issue could deadlock under specific timing. Phase 5/6 unit tests cover the happy path + propagate path + cancel path, but not "all-thunks-completed-with-error-results-but-something-async-is-still-pending."

## Steps to reproduce

Approximate (the live smoke is the original repro):

1. Set up a workflow context where `tabstack_research` (or any other tool) returns an error-shaped `ToolResult`.
2. Run `/research <topic>` with answers; observe the parallel completes (5 children landed) but doesn't return.

A minimum repro test would be valuable — likely in `tests/test_workflow_parallel.py` or `tests/test_workflow_pipeline.py`.

## Priority

[...]

PR #588, opened 2026-06-13
refactor(conversations): remove flat-layout sidecar fallback

lmorchard:
## Summary
- Removes the deprecation-window fallback that let the code read/write the old flat `conversations/{conv_id}.SUFFIX` sidecars in place.
- Now that the per-conversation directory layout (#576/#578) is the only layout and existing data is migrated, the code can have one path instead of two.

## Design Decisions
- **Keep the migration script + `make migrate-sidecars(-dry)` + `SIDECAR_FILENAMES`.** Removing the *runtime* fallback doesn't remove the need to migrate an instance upgrading from the pre-#576 flat layout — the one-shot script is the supported path. Without the fallback, that migration is now a hard prerequisite (no silent in-place reads).
- **Keep `sidecar_path` as a thin chokepoint** (`conversation_dir(...) / filename`) rather than inlining at 9 sites — one documented seam for the convention.
- **Remove the no-split-archive invariant test.** It guarded "append to an existing flat archive in place" — behavior that's intentionally gone.

## Changes
- `conversation_paths.py`: `sidecar_path` drops `legacy_suffix` + fallback; `_legacy_flat_path` removed; `iter_conversation_archives` is dir-only (still excludes compacted, still fails open on `OSError`); `delete_conversation_files` rmtrees the `{id}/` dir, plus a best-effort *delete-only* cleanup of any leftover flat `{id}.SUFFIX` files (defense for a partially-migrated instance — not a read/write fallback); `SIDECAR_LEGACY_SUFFIXES` deleted (unused).
- 9 sidecar call sites drop the `legacy_suffix` arg.
- Tests: fallback / both-layout / no-split tests removed; the rest rewritten to the single-layout contract.
- Docs: `data-layout.md` reframes the flat layout as a one-time migration (no runtime fallback); `CLAUDE.md` key-file note updated.

Net **−158 lines** (a subtraction).

## Migration prerequisite (done for the known instances)
Removing the fallback makes any remaining flat sidecar invisible, so every instance must be migrated first.
- **Deployed agent:** migrated (you ran `make migrate-sidecars`).
- **Local `data/`** (shared by worktrees via `.env`): was *not* migrated — found 6,465 flat files. Backed up + migrated this session (6,465 moved → 0 flat remain; idempotent re-run = 0; archives read back via the dir layout). See `notes.md`.

Any *other* instance must run `make migrate-sidecars` (after `make migrate-sidecars-dry` + a backup) before deploying this.

## Test Plan
[...]

PR #603, opened 2026-06-25
feat(workflow): skill activation — auto-load + @workflow(requires_skills=...) (#580)

### LINKED ISSUE #580: Workflow contexts can't reach skill-bundled tools (e.g., tabstack_research)
## Context

Workflows running under `TurnKind.WORKFLOW` can't reach tools that are bundled inside skills. `execute_tool("tabstack_research", ...)` returns `[error: unknown tool 'tabstack_research'. Did you mean: workspace_search.]` because the tabstack skill isn't activated for the workflow turn — only the agent loop activates skills (via `activate_skill`).

This surfaced during the live smoke for #574 ([PR #579](https://github.com/lmorchard/decafclaw/pull/579), Finding 1). The `/research` hero workflow now fails fast with a clear `RuntimeError` when every search returns an error result, but the underlying gap remains: skill tools are simply unreachable from workflows.

## What `wf.tool_call` does today

`wf.tool_call(name, **args)` calls `execute_tool(self.ctx, name, args)` against `ctx.tools.allowed`. The allowlist is right; the registry is the problem — skill tools aren't loaded into `ctx.tools.extra` unless the owning skill activates.

## Possible directions

- **`@workflow(..., requires_skills=["tabstack"])`** — declare needed skills at decoration time; the engine activates them before invoking the orchestrator. Most explicit; most code.
- **Auto-activate a configurable set of "workflow-safe" skills** when a workflow turn starts (`workflow.auto_activate: ["tabstack", "vault", ...]` in agent config). Simpler API; less per-workflow control.
- **Make skill tools always discoverable from workflows** (pre-activate the catalog). Loosens the skill activation model; broadest change.

Worth a brainstorm before picking.

## References

- Smoke transcript: `docs/dev-sessions/2026-06-10-1732-574-workflow-batch-primitives/smoke.md` (Finding 1)
- PR #579 (merged)
- Workflow engine docs: [docs/workflows.md](../blob/main/docs/workflows.md)

lmorchard:
## Summary

Workflows can now reach skill-bundled tools. Closes the gap surfaced by #574 PR #579 live smoke Finding 1 — `/research` was unable to call `tabstack_research` because workflow turns bypassed the agent loop's skill activation. This branch wires activation into `run_workflow_turn`.

Two halves:

1. **Always-loaded skills** (e.g. `vault`, `background`, `mcp`) auto-activate for every workflow turn, matching the existing USER-turn behavior. Fail-soft per skill so one broken always-loaded skill doesn't take down every workflow.
2. **`@workflow(name, *, requires_skills=("tabstack",))`** lets a workflow declare additional non-always-loaded skills it needs. Fail-loud — a missing skill or `init()` failure surfaces as `WorkflowSkillActivationFailed` BEFORE the orchestrator runs.

`/research` now declares `requires_skills=("tabstack",)` and the live smoke confirmed real `tabstack_research` invocations land 4-6KB of markdown into the journal (vs. `[error: unknown tool]` in the prior smoke).

## Design Decisions

(Full context in `docs/dev-sessions/2026-06-12-1435-580-workflow-skill-activation/spec.md`.)

- **Hybrid: always-loaded auto-activation + per-workflow `requires_skills`.** Rejected always-loaded-only (doesn't unblock tabstack without bloating every agent-loop turn's critical-tool budget) and per-workflow-only (forces every workflow to redeclare vault/background/mcp).
- **Fail-loud for declared skills, fail-soft for always-loaded.** A typo in `requires_skills` becoming a runtime "unknown tool" error 30s into a turn is worse than a setup-time failure. Always-loaded retains the existing USER-turn fail-soft posture so one broken admin skill doesn't take down everything.
- **Reuse `activate_skill_internal`.** No new activation code path. The same machinery that populates `ctx.tools.extra` / `ctx.tools.extra_definitions` / `ctx.skills.activated` for the agent loop populates them for the workflow turn.
- **Silent-`ToolResult` guard.** `activate_skill_internal`'s tool-load-failure path returns a `ToolResult` instead of raising (`tools/skill_tools.py:228-230`). Phase 2's helper checks `ctx.skills.activated` membership AFTER the call and raises `WorkflowSkillActivationFailed` if the skill didn't actually register. Caught in code review; regression test guards it.

## Changes

- `src/decafclaw/skills/__init__.py` — `activate_always_loaded(ctx)` (extracted from `_setup_turn_state`) + new `activate_skills_for_workflow(ctx, names)`.
- `src/decafclaw/workflow/errors.py` — `WorkflowSkillActivationFailed` extending `WorkflowError`.
- `src/decafclaw/workflow/registry.py` — `WorkflowSpec.requires_skills: tuple[str, ...]` + keyword-only `@workflow(... requires_skills=())`.
- `src/decafclaw/workflow/resume.py` — `run_workflow_turn` activates before `run_workflow`; catches `WorkflowSkillActivationFailed` → error `ToolResult` + journal status="error".
- `src/decafclaw/workflow/workflows/research.py` — declares `requires_skills=("tabstack",)`.
- `src/decafclaw/agent.py` — `_setup_turn_state` calls the extracted helper.
- `docs/workflows.md` — new "Skill activation" subsection.

Tests: +18 (2931 → 2949) across `tests/test_skills.py`, `tests/test_workflow_registry.py`, `tests/test_workflow_resume.py`, `tests/test_workflow_research.py`.

## Test Plan

- [x] `make check` clean (ruff + pyright + tsc).
- [x] `make test` → 2949 passed.
- [x] Live smoke: `/research kelp forest restoration` on `vertex-gemini-flash`. Run 1 (no `TABSTACK_API_KEY`): fail-loud activation correctly raised `[error: skill activation failed: requires_skills entry 'tabstack' is not a discovered skill]` with journal status="error". Run 2 (`TABSTACK_API_KEY` set): four `wf.parallel` thunks each completed a real `tabstack_research` invocation; child entries at `(3, 0..3, 0)` carry 4-6KB of markdown per query.
- [x] Activation idempotency on resume (no re-activation of already-activated skills) — `test_run_workflow_turn_activation_idempotent_on_resume`.

Smoke transcript: `docs/dev-sessions/2026-06-12-1435-580-workflow-skill-activation/smoke.md`.

## Acceptance vs scope

**Met:** the `/research` orchestrator reaches a real `tabstack_research` tool — the load-bearing unblock for #580.

**Out of scope (separate issue):** the live walk does NOT complete end-to-end because of a pre-existing `wf.parallel`/`wf.pipeline` primitive bug (#574 smoke Finding 3 → tracked as #582). This smoke materially advances #582's diagnosis: the workflow hangs at the same place with REAL tabstack input as it did with degenerate `[error: …]` input in #574's smoke, so the "Vertex throttling on degenerate input" hypothesis is now ruled out. The bug is in the primitives themselves. I'll update #582 with this evidence after this PR opens.

## References

- Spec: `docs/dev-sessions/2026-06-12-1435-580-workflow-skill-activation/spec.md`
- Plan: `docs/dev-sessions/2026-06-12-1435-580-workflow-skill-activation/plan.md`
- Research: `docs/dev-sessions/2026-06-12-1435-580-workflow-skill-activation/research.md`
- Notes (incl. execute-phase decisions and bug-finds): `docs/dev-sessions/2026-06-12-1435-580-workflow-skill-activation/notes.md`
- Smoke transcript: `docs/dev-sessions/2026-06-12-1435-580-workflow-skill-activation/smoke.md`
- Related: #574 / PR #579 (the surfacing smoke), #582 (the primitive bug that prevents full end-to-end walk).

Closes #580


lmorchard:
Addressed the Copilot inline comment + squashed.

**Copilot fix:** `requires_skills="tabstack"` would silently iterate as characters via `tuple("tabstack")` → `('t','a','b','s','t','a','c','k')`, causing baffling activation failures later. Added an `isinstance(requires_skills, str)` check at decoration time that raises `TypeError` with a clear suggestion (`Did you mean requires_skills=("tabstack",) ?`). Regression test in `test_workflow_decorator_rejects_bare_string_requires_skills`.

**Squashed:** rebased onto the latest origin/main (picked up #588 — flat-layout sidecar removal), then collapsed the 6 phase commits + Copilot fix into a single commit. `make check` clean; 2943 tests pass on the squashed tip.

PR is now one commit, ready for re-review or merge.

Issue #598, opened 2026-06-24
Agent diagnostic-discipline guardrails (loop-breaker, verify-tool-fired, apology spiral)


lmorchard:
## Problem

In conversation `web-lmorchard-fa5ec853` (Caffeina / vertex-gemini-pro), the agent fell into a ~150-message degenerate loop while debugging a skill that wouldn't load. Three distinct behavioral failure modes, all amenable to prompt guardrails:

1. **No loop-breaker.** It repeated the same failed edit-and-refresh cycle dozens of times, guessing at causes without ever switching to diagnosis (read logs, minimal repro, check the contract). Want: after ~2 identical failed attempts, STOP editing and switch to root-cause diagnosis.

2. **Phantom tool call.** At one point it wrote out a full plan to re-engage Claude Code but **never emitted the `claude_code_send` call** — the user had to catch it ("I don't think you actually engaged with claude code"). Want: after describing an action, verify the tool call actually fired before reporting progress.

3. **Apology spiral.** Nearly every turn opened with "You are absolutely correct, my apologies, this is unacceptable…" then repeated the same failed move. Pure token waste, false signal of learning. This is a known degenerate Gemini-Pro register. Want: discourage performative self-flagellation; channel into a concrete changed approach or stop.

## Notes

- Token budget was NOT the cause (87K / 300K window). This was diagnostic-discipline collapse.
- These are agent-persona / system-prompt guardrails, possibly model-conditional (Gemini Pro is the worst offender here).
- Consider an eval case: a deliberately-unloadable skill, assert the agent pivots to log/contract inspection within N tool calls rather than thrashing edits.

_Filed from a session postmortem._

Issue #597, opened 2026-06-24
Surface the workspace-skill authoring contract at skill-creation time


lmorchard:
## Problem

The contract for authoring a workspace skill (SKILL.md frontmatter shape, `tools.py` filename, `get_tools(ctx) -> (dict, list)` signature, how tools.py is imported, that `default_api.*` is not a thing) is documented in `docs/skills.md` and CLAUDE.md but is **not present in the context** when the agent is actually authoring a skill via the `claude_code` skill or freehand.

## Proposed

When the agent is creating a workspace skill, surface a concise authoring contract — either:
- A short reference block injected by the skill-creation / `claude_code` path, or
- A bundled `skill-creator`-style skill body the agent activates, that states the minimal SKILL.md + tools.py template.

Minimal correct template:

```
skills/<name>/
  SKILL.md      # --- frontmatter with name + description --- then markdown body
  tools.py      # def get_tools(ctx) -> (dict, list): ...
```

## Context

Incident in #595/#596: the agent had no in-context knowledge of the frontmatter requirement or the `tools.py` filename, and cargo-culted a Gemini-function-calling `default_api.*` API that doesn't exist in decafclaw.

_Filed from a session postmortem._

PR #608, opened 2026-06-26
feat(skills): rejection reporting + skill_validate lint (#595, #596)

### LINKED ISSUE #595: refresh_skills should report rejected skills and the reason
## Problem

When a workspace skill fails validation during discovery, it is **silently dropped**. `skills/__init__.py:_parse_skill_md` logs a `log.warning(...)` (e.g. `"No valid YAML frontmatter in {path}"`, `"Missing 'name'"`, `"Missing 'description'"`) and returns `None`. The `refresh_skills` tool then returns only the **successful** skill list — the agent gets no signal that a skill it just authored was found-but-rejected, or why.

## Impact (real incident)

In conversation `web-lmorchard-fa5ec853` (Caffeina / vertex-gemini-pro), the agent authored `skills/meta-ingest/SKILL.md` with **no YAML frontmatter** (it started directly with a `# Meta-Ingest Skill` markdown heading). The skill was silently rejected on every `refresh_skills`. The agent — never seeing the log — spent **~150 messages** editing the unrelated `main.py`, guessing at SyntaxErrors and sandbox constraints, even reducing `main.py` to a no-op stub (still failed), before giving up. A single rejection reason in the tool result would have ended the session at the first refresh.

## Proposed fix

Have `refresh_skills` return rejected entries alongside the loaded list, e.g.:

```
Skills refreshed. Loaded: [...]
Rejected:
  - skills/meta-ingest/SKILL.md — no valid YAML frontmatter
```

Requires threading rejection reasons out of `discover_skills` / `_parse_skill_md` (currently log-only) into the tool's return value. This is the single highest-leverage fix for the silent-failure trap.

_Filed from a session postmortem._

### LINKED ISSUE #596: Add a skill_validate / lint tool for pre-flight skill authoring
## Problem

There's no way for the agent (or a human) to validate a workspace skill *before* `refresh_skills`. Authoring mistakes only surface as "skill didn't appear" with no diagnosis.

## Proposed

A `skill_validate(path)` tool (or a `--check` mode on refresh) that reports, for a skill directory:

- SKILL.md present and has valid `---` YAML frontmatter with `name` + `description`
- `tools.py` present (if the skill declares tools) — **not** `main.py` or other names
- `tools.py` imports cleanly and exports `get_tools(ctx) -> (dict, list)` with the correct signature
- `allowed-tools` / `requires.env` parse correctly

Returns a checklist of pass/fail with actionable messages. Pairs with #595 (rejection reporting) — this is the proactive pre-flight, that is the reactive post-mortem.

## Context

Same incident as #595: the agent also used `main.py` instead of `tools.py` and wrote `get_tools()` instead of `get_tools(ctx)`, plus referenced a nonexistent `default_api.shell` API. A lint tool catches all of these before the loader does.

_Filed from a session postmortem._

lmorchard:
## Summary

Two related fixes from the `web-lmorchard-fa5ec853` postmortem — an agent burned ~150 messages because a workspace skill it authored (missing YAML frontmatter) was **silently rejected** on every `refresh_skills`, with no signal as to why.

- **#595** — `refresh_skills` now reports skills found-but-rejected during discovery, with the reason, under a `Rejected (found but not loaded):` section.
- **#596** — new `skill_validate(path)` pre-flight lint tool: validates one workspace skill directory deeply and returns a ✓/✗ checklist — frontmatter (`name`/`description`), `tools.py` (not `main.py`), clean `tools.py` import (catches `SyntaxError`/`NameError`/`ImportError`), and a `get_tools(ctx)` signature (introspected, **not** called) or `TOOLS`/`TOOL_DEFINITIONS`.

### Design — single source of truth (the load-bearing decision)

One `validate_skill_md()` is the sole discovery-level accept/reject decision. `parse_skill_md` became a thin `validate → build_skill_info` wrapper (behavior unchanged); `discover_skills` and `skill_validate` both route through `validate_skill_md`. So rejection reasons and lint checks **cannot drift** from the actual accept/reject decision — the exact failure mode the postmortem warned about.

`refresh_skills` surfaces rejections via an optional `rejections` accumulator threaded `discover_skills → load_system_prompt → refresh_skills` — plumbing only, so existing callers/tests are unaffected (no return-shape churn).

## Changes

- `skills/__init__.py` — `validate_skill_md` / `build_skill_info` / `SkillValidation` / `CheckResult` / `SkillRejection`; `parse_skill_md` → wrapper; `discover_skills` gains optional `rejections`.
- `prompts/__init__.py` — `load_system_prompt` threads `rejections` through.
- `tools/skill_tools.py` — `tool_skill_validate` + `_lint_tools_py` + `_render_validation`; registered at `low` priority.
- `docs/skills.md` — rejection reporting + `skill_validate` contract checklist.
- `evals/tool_choice/core_overlaps.yaml` — `skill_validate` vs `refresh_skills` disambiguation case.
- Tests across `tests/test_skills.py` + `tests/test_prompts.py`.

## Test plan

- `make check` green (ruff, pyright 0 errors, message-types drift, tsc).
- Full `make test`: **2961 passed**.
- `make eval-tools`: the new `skill-validate-vs-refresh-preflight` case passes (model chose `skill_validate` over `refresh_skills`).
- Reproduces the incident's failures in unit tests: `SyntaxError`, undefined-name (`default_api`), and zero-arg `get_tools()`.

## Out of scope (same postmortem cluster)

- **#597** — surface the authoring contract in-context at creation time.
- **#598** — agent diagnostic-discipline guardrails + behavior eval.

Session docs: `docs/dev-sessions/2026-06-25-1716-595-596-skill-validation/`.

Closes #595
Closes #596

### DIFF
--- a/docs/dev-sessions/2026-06-25-1716-595-596-skill-validation/notes.md
+++ b/docs/dev-sessions/2026-06-25-1716-595-596-skill-validation/notes.md
@@ -1,0 +1,82 @@
+# Session notes — #595/#596 skill validation
+
+## Summary
+
+Implemented two related fixes from the `web-lmorchard-fa5ec853` postmortem (an agent
+burned ~150 messages because a workspace skill it authored — missing YAML frontmatter —
+was silently rejected on every `refresh_skills`):
+
+- **#595** — `refresh_skills` now reports skills found-but-rejected during discovery, with
+  the reason, under a `Rejected (found but not loaded):` section.
+- **#596** — new `skill_validate(path)` pre-flight lint tool: validates one workspace skill
+  dir deeply (imports `tools.py` to catch `SyntaxError`/`NameError`/`ImportError`,
+  introspects but does not call `get_tools`'s signature, checks `tools.py` vs `main.py`),
+  returns a ✓/✗ checklist as a `ToolResult` with structured `data`.
+
+**Core design (the load-bearing decision):** one `validate_skill_md()` is the single source
+of truth for the discovery-level accept/reject decision. `parse_skill_md` became a thin
+`validate → build_skill_info` wrapper (behavior unchanged); `discover_skills` and
+`skill_validate` both route through `validate_skill_md`. Rejection reasons and lint checks
+therefore cannot drift from the actual accept/reject decision — the exact failure mode the
+postmortem warned about.
+
+## What shipped
+
+| Commit | What |
+|--------|------|
+| `31301df` | refactor: extract `validate_skill_md` / `build_skill_info`; `parse_skill_md` → wrapper |
+| `489a95e` | feat: `refresh_skills` reports rejections (#595) via optional `rejections` accumulator threaded through `discover_skills` → `load_system_prompt` |
+| `fe33ee8` | feat: `skill_validate` tool (#596) + `docs/skills.md` |
+| `17ccd2d` | test: assert `skill_validate` no-ctx message mentions ctx |
+| `8546589` | test(evals): `tool_choice` case `skill_validate` vs `refresh_skills` |
+| `c29ec62` | test(prompts): update `discover_skills` monkeypatch fakes for the new `rejections` kwarg |
+
+Plus the spec/plan doc commits (`f69154e`, `cd2ebe0`).
+
+## Verification
+
+- `make check` green (ruff, pyright 0 errors, message-types drift, tsc).
+- Full `make test`: **2942 passed**.
+- `make eval-tools`: the new `skill-validate-vs-refresh-preflight` case passed (model chose
+  `skill_validate` over `refresh_skills`, no overlap). 3 unrelated pre-existing eval failures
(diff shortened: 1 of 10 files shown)


lmorchard:
Addressed both Copilot findings:

- **Frontmatter message** (`skills/__init__.py`) — broadened to cover malformed-YAML-inside-delimiters and missing/mismatched `---` delimiters, not just "must start with a block." New regression test `test_validate_malformed_yaml_is_frontmatter_failure`.
- **`_rejection_display_path` leak** (`skill_tools.py`) — paths outside the workspace/agent roots (absolute `extra_skill_paths`) are now redacted to the trailing `<skill-dir>/SKILL.md` segments instead of echoing the full host path. New tests `test_rejection_display_path_relative_to_workspace` / `_redacts_external_absolute`.

Rebased onto latest `main` (resolved conflicts with #603), verified `make check` + full `make test` (2964 passed) + the new `tool_choice` eval case, then squashed to a single commit.

PR #606, opened 2026-06-26
feat(blog-ideas): daily blog-idea brainstorm contrib skill

lmorchard:
## What

A new **opt-in contrib skill** (`contrib/skills/blog-ideas/`) that, once a day, reviews the week-so-far across everything I've been reading, writing, and thinking, and maintains a living tiered "blog ideas this week" vault page.

- **Inputs** (all already in the synced vault — zero new ingest plumbing): the agent's ingest folders (`agent/pages/{bookmarks,mastodon,youtube,github,podcasts,music,kindle}`), the daily Obsidian journal (`journals/`, all sections — `*(blog candidate)*` a stronger-but-not-exclusive signal), and the blog archive (`blog/drafts` to resurface stalled drafts; `blog/daily`/`weeknotes` for dedup, follow-ups, and voice).
- **Output:** a living weekly page at `agent/pages/blog-ideas/{ISO-week}.md`, rewritten in place each day — tiered **Headliners** (3–5) + **Seeds** + **Recurring** (cross-week themes surfaced as signal, not re-pitched). Under `agent/` to avoid the out-of-agent write gate a scheduled run can't satisfy.
- **Delivery:** rides the existing newsletter's scheduled-activity aggregation — **no newsletter changes**. The run's narrative summary is the newsletter's raw material and the nudge to read the page.

## Design notes

- **Almost prompt-only** (`SKILL.md` + `SCHEDULE.md`), mirroring `dream`/`garden`, **plus one small `tools.py`**: a deterministic `blog_ideas_week(offset_weeks=0)` helper. `current_time` has no ISO-week number, and the living-page model needs a stable week key (a wrong key duplicates the page instead of refining it). ISO-week math / year boundaries belong in code, not the LLM.
- **Single-turn `dream`-style synthesis**, not fan-out or the workflow engine — the value is cross-source connection, which wants one reasoning context.
- **Scoped to the scheduled path** (+ the explicit `/blog-ideas` command). A mid-implementation experiment to route casual natural-language questions was **reverted** (it over-loaded the skill and risked the scheduled prompt being misclassified and silently no-op'ing); the cross-cutting `context_composer.py` change it introduced is spun out as #604.

## Contrib (opt-in) — activation

This is a personal, vault-layout-specific skill, so it lives in `contrib/`, not as a core bundled skill. To activate on a deployment:
1. Add `$CONTRIB/skills/blog-ideas` to the agent's `extra_skill_paths`.
[...]

PR #610, opened 2026-06-26
fix(workflow): wf.parallel/wf.pipeline hang on FIRST_EXCEPTION fallback (#582)

### LINKED ISSUE #582: Investigate 3-minute hang in wf.pipeline summarize stage against error-text input

lmorchard:
## Summary

`wf.parallel` and `wf.pipeline` hung forever in production when:
1. All thunks/items completed normally (no exception),
2. `ctx.cancelled` was a non-None `asyncio.Event` (so the cancel_watcher existed),
3. `ctx.cancelled.set()` was never called during the wait.

**Root cause.** `asyncio.wait(wait_set, FIRST_EXCEPTION)` falls back to `ALL_COMPLETED` when no future raises (per Python docs). The cancel_watcher (awaiting an `asyncio.Event` that never fires) kept the wait hung forever once all real thunks completed. Unit tests didn't catch it because mock-ctx fixtures set `ctx.cancelled = None`, skipping the watcher entirely; live workflows always have a real Event (`ConversationManager` constructs one per turn), so the bug fired in production the moment any workflow's `wf.parallel`/`wf.pipeline` thunks didn't raise.

## The fix

Race `asyncio.gather(*tasks)` against the cancel_watcher with `FIRST_COMPLETED` instead. The `_GatheringFuture` (the gather's return value, used directly — not wrapped in `asyncio.create_task`) completes the moment the last thunk returns OR the first raises. The cancel_watcher fires only on real cancellation. Either way, exactly one future enters `done` and the wait exits.

Same fix in both primitives (deliberate code duplication — two call sites is below the third-callsite-extraction threshold).

## Design notes

- **`return_exceptions=False` is load-bearing.** With `True`, gather waits for ALL inner tasks regardless — a fast_raiser + slow_hanger pair would deadlock the gather. Without `True`, gather completes the moment one raises and cancels the rest (per Python docs). Explicit straggler-cancel + drain before `.result()` ensures cancelled stragglers finish their cancellation handling.
- **`asyncio.gather(...)` is NOT wrapped in `asyncio.create_task`.** Gather returns a `_GatheringFuture` already scheduled on the loop; Python 3.11+ rejects `create_task(future)` since it expects a coroutine. The Future goes directly into `asyncio.wait`'s wait_set.
- **Preserves the `_CancelSignal` + cleanup pattern** from PR #573/#574. The sentinel still distinguishes "cancel fired" from "gather completed" in the `done` set.
- **Preserves PR #579's exception-unmasking semantics.** Previously a `for t in tasks: ... first_exc` loop explicitly skipped CancelledError to surface the REAL exception. Now `gather_future.result()` raises the first-exception-in-time directly — straggler cancellations happen AFTER gather has locked in its exception, so they can't mask the real failure.

## Test Plan

- [x] Two new regression tests directly exercise the bug, both wrapped in `asyncio.wait_for(..., timeout=2.0)` to fail loudly via TimeoutError on regression:
  - `test_workflow_parallel.py::test_parallel_returns_when_all_thunks_complete_with_cancel_event_set_but_not_fired`
  - `test_workflow_pipeline.py::test_pipeline_returns_when_all_items_complete_with_cancel_event_set_but_not_fired`
- [x] Pre-fix: both new tests timed out at 2.0s. Post-fix: both pass in <100ms.
- [x] Existing critical tests STILL pass:
  - `test_parallel_surfaces_real_exception_over_cleanup_cancel` (PR #579 index-vs-time-order regression)
  - `test_parallel_cancels_inflight_on_ctx_cancelled` (mid-run cancel)
  - `test_pipeline_propagates_real_exception_over_cleanup_cancel` (pipeline analog)
  - `test_pipeline_cancels_inflight_on_ctx_cancelled` (pipeline mid-run cancel)
- [x] `make check` clean.
- [x] `make test` → 2972 passing.
- [x] Live smoke walked `/research kelp forest restoration` on `vertex-gemini-flash`. **The outer parallel entry at seq `"3"` LANDS in the post-execution journal** (was missing in #574/#580 smokes — the smoking gun of the bug). Full transcript: `docs/dev-sessions/2026-06-25-1723-582-pipeline-hang/smoke.md`.

## Out of scope (separate follow-up)

The live smoke surfaced an incidental `tabstack_research` 180s-timeout issue (the tool's iterative-research workflow exceeds the per-tool `TOOL_TIMEOUT_SEC` default). The `/research` orchestrator's fail-fast guard correctly caught the all-error result set and raised `RuntimeError`; the workflow exited cleanly with `status="error"` rather than hanging. That's the system working as designed — separate ticket worth filing for `/research`'s tool-timeout or fan-out target.

## References

- Spec: `docs/dev-sessions/2026-06-25-1723-582-pipeline-hang/spec.md`
- Plan: `docs/dev-sessions/2026-06-25-1723-582-pipeline-hang/plan.md`
- Research (including the 8-line standalone repro that confirmed the asyncio semantics): `docs/dev-sessions/2026-06-25-1723-582-pipeline-hang/research.md`
- Notes (incl. execute-phase decisions and the two spec bugs caught mid-execute): `docs/dev-sessions/2026-06-25-1723-582-pipeline-hang/notes.md`
- Smoke transcript + journal snapshot: `smoke.md` / `smoke-journal-snapshot.json`
- Related: #574 / PR #579 (the workflow primitives this fixes), #580 / PR #603 (skill activation — its smoke also hit this hang).

Closes #582

### DIFF
--- a/docs/dev-sessions/2026-06-25-1723-582-pipeline-hang/asyncio-verification.py
+++ b/docs/dev-sessions/2026-06-25-1723-582-pipeline-hang/asyncio-verification.py
@@ -1,0 +1,84 @@
+"""Verify the asyncio semantics the #582 fix depends on.
+
+If any of these assertions fail, the fix shape in spec.md is wrong and
+we need to revisit before touching production code.
+"""
+import asyncio
+
+
+async def quick(i):
+    await asyncio.sleep(0.01)
+    return i
+
+
+async def boom():
+    await asyncio.sleep(0.01)
+    raise ValueError("boom")
+
+
+async def never_fires():
+    ev = asyncio.Event()
+    await ev.wait()
+    raise RuntimeError("unreachable")
+
+
+async def main():
+    # (1) gather(return_exceptions=True) returns BaseException in place of failures.
+    tasks = [
+        asyncio.create_task(quick(0)),
+        asyncio.create_task(boom()),
+        asyncio.create_task(quick(2)),
+    ]
+    result = await asyncio.gather(*tasks, return_exceptions=True)
+    assert result[0] == 0, f"got {result[0]!r}"
+    assert isinstance(result[1], ValueError), f"got {result[1]!r}"
+    assert result[2] == 2, f"got {result[2]!r}"
+    print("(1) gather return_exceptions: OK")
+
+    # (2) Cancelling gather_future cancels all inner tasks.
+    # NOTE: asyncio.gather() returns a _GatheringFuture, NOT a coroutine.
+    # Python 3.11+ rejects asyncio.create_task(gather(...)). Use the gather
+    # return value directly as the Future — same shape used in production.
+    inner_tasks = [
+        asyncio.create_task(asyncio.sleep(10, result=i)) for i in range(3)
+    ]
+    gather_future = asyncio.gather(*inner_tasks, return_exceptions=True)
+    await asyncio.sleep(0)  # let everything start
+    gather_future.cancel()
+    try:
+        await gather_future
+    except asyncio.CancelledError:
+        pass
+    for i, t in enumerate(inner_tasks):
+        assert t.cancelled(), f"inner task {i} not cancelled: done={t.done()}"
+    print("(2) gather.cancel propagates: OK")
+
+    # (3) FIRST_COMPLETED race: returns when gather completes, watcher pending.
+    real_tasks = [asyncio.create_task(quick(i)) for i in range(3)]
+    gather_future = asyncio.gather(*real_tasks, return_exceptions=True)
+    watcher = asyncio.create_task(never_fires())
+    try:
+        done, pending = await asyncio.wait_for(
+            asyncio.wait([gather_future, watcher],
+                         return_when=asyncio.FIRST_COMPLETED),
+            timeout=2.0,
+        )
+    except asyncio.TimeoutError:
+        raise AssertionError("FIRST_COMPLETED race hung — fix shape is wrong")
+    assert gather_future in done, "gather should be done"
+    assert watcher in pending, "watcher should be pending"
+    print("(3) FIRST_COMPLETED race: OK")
+    watcher.cancel()
+    try:
+        await watcher
+    except asyncio.CancelledError:
+        pass
+
+    # (4) gather_future.result() returns list on normal completion.
+    assert gather_future.result() == [0, 1, 2]
+    print("(4) gather_future.result() returns list: OK")
+
+
+if __name__ == "__main__":
+    asyncio.run(main())
+    print("\nAll assertions passed. Fix shape from spec.md is sound.")
--- a/docs/dev-sessions/2026-06-25-1723-582-pipeline-hang/notes.md
+++ b/docs/dev-sessions/2026-06-25-1723-582-pipeline-hang/notes.md
@@ -1,0 +1,55 @@
+# Notes — Pipeline hang investigation (#582)
+
+## Session start — 2026-06-25 17:23
+
+- Branch: `fix/582-pipeline-hang` (from `origin/main` @ `9deab85`)
+- Worktree: `/Users/lorchard/devel/decafclaw/.claude/worktrees/fix-582-pipeline-hang`
+- Session dir: `docs/dev-sessions/2026-06-25-1723-582-pipeline-hang/`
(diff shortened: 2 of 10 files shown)


lmorchard:
Addressed all 5 Copilot inline comments + squashed.

**Copilot's findings — all stale `return_exceptions=True` references in spec.md and plan.md that didn't get cleaned up during the mid-execute pivot.** The production code was correct (`return_exceptions=False` since Phase 2), but I left stale references in the planning docs. All five locations updated:

- `spec.md:3` (Goal line) — now reads `asyncio.gather(*tasks)` (default `return_exceptions=False` — load-bearing).
- `spec.md:123` (Design decision A) — updated rationale; explicitly calls out why `True` would re-introduce a hang.
- `spec.md:131` (was the "preserve unmasking iteration" decision) — rewritten as "Let `gather_future.result()` re-raise the first exception in time; cancel stragglers explicitly before calling it." Matches actual code.
- `plan.md:3` (Goal line) — mirrors spec.md fix.
- `plan.md:175` (code snippet) — drops `return_exceptions=True` from the `gather_future` construction; adds inline comment about the load-bearing nature.

Remaining `return_exceptions=True` references audited and left in place — they're legitimate (verification script's assertion (1) which exercises the `True` semantics, and the production straggler-drain at `spec.md:86` which legitimately wants to absorb cancelled-task exceptions).

**Squashed:** rebased onto current origin/main (picked up #606 / #608 / #609 — blog-ideas skill, skill validation, and the vault-write-inline fix), then collapsed the 4 phase commits + Copilot fixes into a single commit at `61d4722`. `make check` clean; 2974 tests pass on the squashed tip.

PR is one commit, ready for merge.

PR #616, opened 2026-07-06
fix(workflow): /research completes end-to-end — tabstack timeout + narrower fan-out (#613)

lmorchard:
## Summary

Makes `/research` actually usable as a demo of #574's fan-out primitives. Before this fix, all 3-5 parallel `tabstack_research` calls timed out at the 180s per-tool default while making forward progress — the `/research` orchestrator's fail-fast guard then raised `RuntimeError` and the workflow exited with `status="error"` (see #613 for the pre-fix repro).

Post-fix: `/research kelp forest restoration` completes in ~3 minutes on `vertex-gemini-flash` with `status="done"` and a real report dict returned to the client.

## The fix

Two coordinated changes:

1. **`tabstack_research`'s definition carries `"timeout": 600`.** 10-minute defensive ceiling. Tabstack's iterative research (multi-query search + page analysis across 3 iterations) typically takes 2-3 minutes per call; the 180s default fires mid-flight. 600s gives ~3-5× headroom for typical runs while bounding runaway iteration bugs at 10 minutes.

2. **`/research`'s plan-stage generates 2-3 queries instead of 3-5.** Three coordinated edits (schema `minItems`/`maxItems`, `_SYS_PLAN`, `_research_plan_prompt`). Tabstack expands each query into ~7 internal searches per iteration, so 2-3 parallel calls give plenty of breadth without the redundancy of 5.

Regression test in `tests/test_tool_timeout.py` guards the tabstack timeout override from silent removal in future refactors.

## Live smoke evidence

Walked `/research kelp forest restoration` on Flash. **`status="done"` at T+180s.** Final journal:

```
seqs:  0, 1, 2, 3, 3.0.0, 3.1.0, 4, 4.0.0, 4.1.0, 5
kinds: user_input × 2, llm_call, parallel, tool_call × 2, pipeline, llm_call × 2, subagent
```

Both `tabstack_research` calls returned real markdown (5,260 + 4,012 chars — no `[error: timed out]`). Final subagent report titled "Restoring Our Underwater Forests: The Science and Value of Kelp Forest Restoration."

Contrast the #613 pre-fix smoke (`docs/dev-sessions/2026-06-25-1723-582-pipeline-hang/smoke.md`): `status="error"`, all `tool_call` results were `[error: tool tabstack_research timed out after 180s]`, workflow exited via `/research`'s fail-fast guard.

## Design notes

[...]

PR #617, opened 2026-07-07
fix(workflow): auto-resume interrupted status=running journals on startup

### LINKED ISSUE #581: Workflow journals with status=running are stuck on server restart
## Context

If a server crash (SIGINT, OOM, deploy) interrupts a workflow turn mid-LLM, the journal persists at `status="running"` — but on server restart, nothing scans for in-flight workflows and re-enqueues them. The workflow stays stuck until a user action triggers it.

Surfaced during the live smoke for #574 ([PR #579](https://github.com/lmorchard/decafclaw/pull/579), Finding 2). The smoke captured a mid-fan-out crash with the journal preserved correctly (4 sub-handle children at `(3, 0..3, 0)`, outer `(3,)` absent). The replay machinery itself handles this state correctly — unit tests in `tests/test_workflow_parallel.py` and `tests/test_workflow_pipeline.py` cover full-cache + mid-fan-out resume against synthetic journals.

## Why it's stuck

Looking at `src/decafclaw/workflow/resume.py`:

- `run_workflow_turn` is only called from `ConversationManager._start_turn` for `TurnKind.WORKFLOW` dispatch.
- The only enqueue path for `TurnKind.WORKFLOW` is `WorkflowUserInputHandler.on_approve` — i.e., a user responding to a confirmation.
- A `status="running"` journal on disk has no trigger to re-enter the engine.

So `status="suspended"` (waiting on a user_input confirmation) recovers fine via the confirmation-recovery scan; `status="running"` (mid-LLM crash) doesn't.

## Possible directions

- **Startup scan** that loads each conversation's `workflow.json` and re-enqueues `status="running"` turns with `kind=TurnKind.WORKFLOW, metadata={"workflow_name": ..., "resume": True}`. Simplest; safe because the replay machinery already short-circuits cached entries.
- **Client-side "resume workflow" command** — adds a UI affordance for stuck workflows. Useful even if (a) lands, but doesn't fix the unattended case.

(a) is the higher-leverage change.

## Caveats

- Re-enqueue policy needs a cap to avoid replay-storms if the same workflow keeps crashing.
- Workflows that crashed because of a deterministic orchestrator bug (e.g., `WorkflowNonDeterministic`) shouldn't re-enqueue — `status="error"` already excludes them, but the policy should be explicit.

## References

- Smoke transcript: `docs/dev-sessions/2026-06-10-1732-574-workflow-batch-primitives/smoke.md` (Finding 2)
- Replay engine: `src/decafclaw/workflow/engine.py`, `src/decafclaw/workflow/resume.py`
- PR #573 (engine), PR #579 (primitives)

lmorchard:
Closes #581.

## Summary

If a server crash (SIGINT, OOM, deploy) interrupts a workflow turn mid-LLM, the journal persists at `status=\"running\"` — but on restart nothing re-enters the engine. The workflow stays stuck until a user action triggers it. Suspend/resume via user_input already recovered correctly through the confirmation-recovery scan; mid-execution crashes did not.

This adds a sibling scan `ConversationManager.startup_scan_workflows()` that walks each conversation, finds `workflow.json` with `status=\"running\"`, increments a persisted attempt counter, and re-enqueues the workflow turn. The replay engine short-circuits already-cached primitives, so the resumed turn picks up exactly where the crash left off.

- **Attempt cap** — bounded by `config.workflow.max_resume_attempts` (default 3, env `WORKFLOW_MAX_RESUME_ATTEMPTS`). At the cap, the journal is marked `status=\"error\"` and skipped, preventing a deterministically-crashing workflow from replay-storming.
- **Bump-before-enqueue** — the increment is persisted BEFORE `enqueue_turn` runs, so a crash inside the resumed turn still consumes an attempt.
- **Fail-open** — a broken journal in one conversation logs a warning and continues; startup is never blocked by workflow state.

Last of the #574 smoke follow-ups. Companion to #573 (engine), #579 (primitives), #603 / #610 / #616 (other smoke gaps).

## Design (five phases on branch, will squash on merge)

| Commit | Change |
|---|---|
| `e0683a6` + `17959b6` | `Journal.attempts: int = 0` field + round-trip tests |
| `e6227f1` | `WorkflowConfig.max_resume_attempts: int = 3` + env override |
| `2296568` | `startup_scan_workflows()` method + `runner.py` wiring + 8 unit tests |
| `f836775` | E2E integration test proving resume-to-done, no double-execution |
| `06e2ac7` | `docs/workflows.md` — new \"Auto-resume on startup\" section |

Per-phase spec-compliance + code-quality review under subagent-driven-development. Full session notes in `docs/dev-sessions/2026-07-07-1548-581-workflow-autoresume/`.

## Test plan

- [x] `make check` clean (ruff + pyright + tsc + message-types drift)
- [x] `make test` — 2998 passing (+12 vs baseline: 3 journal + 3 config + 8 scan + 1 E2E — minus dedup)
- [x] E2E test in `test_workflow_turn_integration.py` proves resumed workflow reaches `status=\"done\"` without re-invoking cached primitives
- [ ] Manual smoke: start `/research` or `/interview`, `Ctrl-C` mid-turn, `make dev` back up, confirm workflow finishes without user action

## Not doing (deferred)

- Client-side \"resume workflow\" command (nice-to-have; not blocking auto-resume)
- Attempt-counter reset on `suspended → running` transitions (see spec §\"Attempt counter reset\")
- Cross-machine coordination (single-instance deployment)

### DIFF
--- a/docs/dev-sessions/2026-07-07-1548-581-workflow-autoresume/notes.md
+++ b/docs/dev-sessions/2026-07-07-1548-581-workflow-autoresume/notes.md
@@ -1,0 +1,19 @@
+# Notes — Workflow status=running auto-resume (#581)
+
+## Session start — 2026-07-07 15:48
+
+- Branch: `fix/581-workflow-autoresume` (from `origin/main` @ `293b624`)
+- Worktree: `/Users/lorchard/devel/decafclaw/.claude/worktrees/fix-581-workflow-autoresume`
+- Session dir: `docs/dev-sessions/2026-07-07-1548-581-workflow-autoresume/`
+- HTTP_PORT: 18896; `TABSTACK_API_KEY` enabled per prior sessions' learnings
+- Baseline: `make test` — 2983 passing in 19.59s
+
+## Origin
+
+Filed as follow-up to #574's smoke (PR #579 Finding 2). If a server crash interrupts a workflow turn mid-LLM, `workflow.json` persists at `status="running"` but nothing re-enqueues it on server restart. The replay machinery itself is correct (unit tests already cover full-cache + mid-fan-out resume against synthetic journals) — the wiring gap is at the harness layer.
+
+This is the LAST remaining follow-up from #574's smoke trio. After this, the workflow surface should be stable:
+- #580 (skill activation from workflows) — closed
+- #582 (parallel/pipeline hang) — closed
+- #613 (/research timeout) — closed
+- #581 (auto-resume on restart) — this session
--- a/docs/dev-sessions/2026-07-07-1548-581-workflow-autoresume/plan.md
+++ b/docs/dev-sessions/2026-07-07-1548-581-workflow-autoresume/plan.md
@@ -1,0 +1,186 @@
+# Plan — Auto-resume `status="running"` workflows (#581)
+
+Spec: `spec.md`. Research: `research.md`.
+
+Execution mode: TDD per phase where practical (test-first for behavior changes, straight edits for pure data-model additions). Each phase is a single logical commit and independently reviewable.
+
+## Phase 1 — Journal `attempts` field
+
+**Files:**
+- `src/decafclaw/workflow/journal.py`
+- `tests/test_workflow_journal.py` (extend if exists; create otherwise)
+
+**Changes:**
+1. Add `attempts: int = 0` to `Journal` dataclass (line ~45).
+2. Add `"attempts": self.attempts` to `to_dict()` output.
+3. In `from_dict()`, read via `d.get("attempts", 0)`.
+
+**Tests (write first):**
+- `test_journal_attempts_defaults_to_zero` — construct a `Journal`, assert `attempts == 0`.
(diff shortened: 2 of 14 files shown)


lmorchard:
Addressed all three Copilot findings in `67ab0a3`:

1. **`startup_scan_workflows` fail-open** — wrapped `enqueue_turn` in try/except that logs a warning and continues to the next conversation. `attempts` is still persisted before the enqueue call, so the crash-safety guarantee holds unchanged. Updated `test_startup_scan_workflows_increments_before_enqueue` to assert the new behavior (no raise, warning captured, `attempts=1` on disk).
2. **`WorkflowConfig` docstring** — rewrote to name the actual target (`status="running"`), not "suspended".
3. **`Journal.attempts` inline comment** — clarified that the scan bumps only on auto-resume (not "each run"), and flips to error when attempts *reach* the cap (not "exceed", since the check is `>=` and happens before the bump).

Suite still 2998 green, `make check` clean.

Issue #622, opened 2026-07-23
Roadmap: 2026-07-23 backlog triage — direction & Ready column


lmorchard:
Living status read for the direction set by the **2026-07-23 backlog triage** (11-agent fan-out; open issues pruned 208→181). Full reasoning — close/merge rationale, epic groupings, per-cluster findings — is the source of truth in the repo:

`docs/dev-sessions/2026-07-23-0948-issue-triage/triage.md`

_(This issue is trimmed to current status; earlier dated progress logs collapsed into the snapshot below.)_

## Direction: measure before building
Reflection, tool-scoring, and prompt handling were argued from impression, not data. Cheap instrumentation answers "does this earn its keep" before the expensive gated follow-ups begin.

## Arc status
1. **Instrumentation (measure-first) — ✅ done.** #310 tool-usage report, #409 reflection telemetry, #480 Phase 1 prompt-cache (PR #623). `make tool-usage-report` / `make reflection-stats`.
2. **Vault retrieval quality — ✅ done.** #306 recency journal (#640) → #197 self-improving vault: retrieval telemetry + dream/garden frontmatter generation + deterministic importance + backfill CLI + backlink index (#646) → #318 first-class tags: filter search, `vault_tags`, journal + inline `#tags`, Tags tab (#660). `make retrieval-report` / `make backfill-frontmatter`.
3. **Confirmation unification — ⏸ deferred.** #243/#541 closed not-planned; #364 in Backlog. `EndTurnConfirm` already renders on web + Mattermost + terminal via the shared `confirmation_request` event, so unification is cleanliness vs. real regression risk. Revive only if a feature needs confirm-as-widget.
4. **Event-reactive — ◐ mostly dropped.** #96 done; #447 (MCP-serve, once flagged highest-leverage) closed **not-planned**. Only #449 (routines/webhooks) and #553 (structured heartbeat status) survive, in Backlog.

## ⏳ Gated on deploy-and-wait data
Arc-1 telemetry **and** Arc-2 retrieval telemetry both need ~a week of real deployed traffic before meaningful. Do **not** start these blind — they exist to be driven by that data:
- **#307** tool audit / consolidation
- **#591** reflection judges (also needs **#528** eval-suites-by-failure-mode first)
- **#480 Phase 2/3** prompt reorder / `cache_control`

## Ready column (actionable now)
- **#531** eval: surface per-turn context diagnostics
- **#450** pre-agent Python scripts + `[SILENT]` suppression

[...]

Issue #480, opened 2026-05-13
Investigate prompt caching across LLM providers


lmorchard:
## Context

DecafClaw doesn't currently do anything provider-specific for prompt caching, and doesn't measure cache hits either. Across our three providers:

- **OpenAI** (`llm/providers/openai.py`) — automatic caching for prompts ≥1024 tokens on stable prefixes. Response carries `prompt_tokens_details.cached_tokens`; we don't read it.
- **Vertex Gemini** (`llm/providers/vertex.py`) — implicit caching on by default for Gemini 2.5; `usageMetadata.cachedContentTokenCount` is returned but `_VertexStreamState.usage` (`vertex.py:262-266`) only pulls `prompt`/`completion`/`total`.
- **Anthropic via LiteLLM** (`llm/providers/openai_compat.py`) — requires explicit `cache_control: {"type": "ephemeral"}` breakpoints. Not emitted today, so any Claude routing is paying full freight.

The bigger concern than wiring hints is that several DecafClaw patterns mutate the prompt prefix mid-conversation, which invalidates any provider's implicit cache:

- Lazy skill loading (`activate_skill` injects body into prompt)
- Dynamic tool sets per turn (deferred catalog, preempt-search promotions)
- `context_cleanup.clear_old_tool_results` replaces bodies of old tool messages in place
- Compaction rewrites history; decision-slice updates each pass
- Per-turn memory retrieval and `@[[mentions]]` injected via `ContextComposer.compose()`

We don't yet know how much of the prompt is actually re-served vs. re-billed.

## Proposed approach (staged)

**Phase 1 — measure.** Cheap, no behavior change.

- Parse `cachedContentTokenCount` from Vertex `usageMetadata` into the usage dict.
- Parse `prompt_tokens_details.cached_tokens` from OpenAI-compat responses.
- Surface in the per-conversation diagnostics sidecar (`workspace/conversations/{conv_id}.context.json`) and/or `TokenUsage` so the UI popover can show cache hit rate.
- Log a debug line per turn with cached / total prompt tokens.

**Phase 2 — only if measurement shows the cache is shredded.**

- Audit `ContextComposer.compose()` for where volatile per-turn injections sit in the prompt. Move them toward the end so the stable system-prompt + tool-catalog prefix stays cacheable.
- Consider whether `context_cleanup` stub-replacement defeats more cache than it saves in input tokens. (Stubbing rewrites old message bodies → invalidates everything after.)
[...]

PR #623, opened 2026-07-23
feat(telemetry): tool-usage, reflection, and prompt-cache instrumentation

### LINKED ISSUE #310: tools: usage telemetry and unused-tool report
## Motivation

We have 50+ tools and no data on which are actually used. [MCP Overload](https://promptforward.dev/blog/mcp-overload) argues that tool count degrades performance; the first move in any consolidation is knowing which tools are load-bearing and which are decorative. Intuition is a poor guide here — some tools added for one flow may have become redundant after later refactors.

## Goals

- Log every tool call with: timestamp, conversation id, tool name, source (core / skill / MCP server), outcome (success / error / end_turn), duration, rough input/output sizes.
- Aggregate a monthly report: calls per tool, unique conversations per tool, error rate, last-called timestamp.
- Surface unused / low-usage tools as consolidation candidates.
- Lightweight — no external telemetry service; append to a local log file or SQLite table.

## Scope

- New `src/decafclaw/tool_telemetry.py` (or extend an existing sidecar) that subscribes to `tool_start` / `tool_end` events and persists records.
- Storage: `{workspace}/tool_usage.jsonl` (append-only, human-readable) or a SQLite table if we want query ergonomics.
- Config: `config.telemetry.tool_usage_enabled`, `config.telemetry.tool_usage_path`.
- Makefile target `make tool-usage-report` that produces a ranked markdown/text report from the log.
- Privacy-sensitive fields (tool args, returns) explicitly NOT logged — only metadata.

## Open questions

- JSONL vs. SQLite — JSONL fits the "files on disk, human-readable" convention; SQLite is easier to query over long windows. Probably JSONL with an optional SQLite index built on demand.
- Retention / rotation — keep forever? Rotate monthly?
- Does this belong in `workspace/` (per-agent, persistent) or somewhere more ephemeral?
- Should we also log failed tool lookups (model called a tool we don't have) to catch deferred-catalog gaps?

## Related

- `src/decafclaw/events.py` (event bus for subscribing)
- `src/decafclaw/agent.py` (publishes `tool_start` / `tool_end`)
- #303 (tool-choice eval — can use this data as ground truth for real-world ambiguity)
- First tool-audit issue in this batch (consumes the monthly report)
- [MCP Overload article](https://promptforward.dev/blog/mcp-overload)

### LINKED ISSUE #409: Reflection: gather telemetry on cost vs effectiveness — does it actually self-correct?
## Problem

Anecdotally, reflection (`reflection.py`) feels expensive relative to what it produces. The two outcomes I see most often are:

1. **Pass-through.** Judge approves on the first call. The judge LLM call ran, paid tokens, didn't change the response. Pure overhead.
2. **Rejection loop.** Judge rejects iteratively, the agent retries, the judge rejects again — usually because the rejection itself is the bug (judge misreading tool results, misinterpreting context, demanding evidence that's already present). Multiplies tokens AND user wait, and the user sees the *last* attempt which is sometimes worse than the first.

Genuine valuable self-correction — agent fixes a real mistake mid-turn because of the judge — feels rare.

Before deciding whether to keep, gate, or kill reflection, we should **collect data** rather than continue arguing from impression.

## What we have today

- One log line per eval: `Reflection result: passed=… critique=… error=…` (`agent.py:270`).
- One `ctx.publish(\"reflection_result\", …)` event per eval.
- Per-call max retries (`config.reflection.max_retries=2`).
- No aggregate counters, no token accounting, no \"did the retry meaningfully change anything\" measurement.

## Proposed telemetry

Add a small reflection-metrics sink (sidecar JSONL under `workspace/reflection/` or rolled into the existing diagnostics sidecar) that records, per turn:

- **Outcome bucket**: `passed_first` / `passed_after_retry` / `loop_exhausted` / `errored` / `skipped` (with reason: `child_agent`, `cancelled`, `empty`, etc.).
- **Retry count**: how many judge rounds before exit.
- **Token cost**: judge prompt + completion tokens per round; sum across rounds.
- **Response delta when retried**: simple measure — character-length change, plus a token-overlap ratio between first and final response. \"Did the retry actually rewrite anything?\" If retries produce near-identical output, the loop is performative.
- **Critique fingerprint when retried**: first ~120 chars of the rejection critique, so we can tell whether the same critique pattern keeps repeating across loops.

Optional but cheap to add at the same time:

- Cumulative counters surfaced via `/health` or a new `/api/reflection/stats` (last N hours: pass-rate, mean retries, mean tokens, loop-exhausted-rate).
- Aggregate roll-up emitted as a notification or vault page on a slow cadence (daily).

## Why this is worth it

Once we have a week of data we can answer:

1. What fraction of turns end in `passed_first`? (Pure overhead measure.)
2. What fraction end in `loop_exhausted`? (Waste + bad UX measure.)
3. When retries succeed (`passed_after_retry`), do the responses meaningfully differ? (Genuine-value measure.)
4. What's the actual token cost of reflection per active hour?

That answers \"is this earning its keep\" with numbers instead of impressions, and lets us make follow-up decisions:

- Disable reflection by default and gate it on a per-conversation flag.
- Add a no-op pass-through path for turn shapes where reflection rarely fires usefully (e.g. tool-heavy turns).
- Lower `max_retries` if the data shows retry-2 almost never converges.
- Or: leave it on, with confidence.

## Out of scope

- Changing reflection behavior. This issue is purely about instrumentation.
- The judge prompt itself. If the data shows the judge is the problem, that's a separate investigation.
- The eval-runner's `reflect_on_failure` (different concern).

## Files to touch

- `src/decafclaw/reflection.py` — add a structured-result emitter
- `src/decafclaw/agent.py` — record per-turn outcome + token cost from `ctx.tokens`
- New: `src/decafclaw/reflection_metrics.py` (or inline) — rolling JSONL writer + simple aggregation
- New: `docs/reflection.md` section on the metrics + how to read them
- New: `tests/test_reflection_metrics.py`

lmorchard:
## Summary
- Adds measure-before-building instrumentation for the three most-argued subsystems — the tool registry (#310), reflection (#409), and prompt handling (#480) — surfaced by the 2026-07-23 backlog triage as the strongest cross-cutting signal: these are reasoned about from impression, and their expensive follow-ups are correctly *gated* on data that doesn't exist yet.
- **Changes no behavior.** Pure data-collection surfaces plus a way to read them. Decisions about what to do with the data are out of scope and become follow-up issues once ~a week of real numbers exists.

## Design Decisions
- **Shared shape** across all three: append-only JSONL sidecars under `workspace/`, a new `TelemetryConfig` group (defaults→config.json→env), fail-open EventBus-subscriber producers, reports via `make` targets.
- **Metadata only** — never tool args/returns, reflection response bodies, or prompt contents; only names, sizes, counts, token totals, deltas, and a 120-char critique fingerprint.
- **Enabled by default** so a deployed agent starts collecting immediately (the point is a week of real data). No log rotation yet — append-only; retention is a follow-up.
- **Two forks confirmed with the author during brainstorm:** (1) enrich `tool_end` events with `conv_id`/`duration_ms`/`input_bytes` at the publish site rather than pairing start/end in the subscriber; (2) emit a reflection row only for *judge-eligible* turns (enabled, non-child, reached a no-tool final), keeping pass-rate denominators meaningful.

## Changes
- **#480 Phase 1 — prompt-cache tokens:** providers normalize a common `cached_tokens` usage key (Vertex `cachedContentTokenCount`; OpenAI-compat `prompt_tokens_details.cached_tokens`); `TokenUsage.total/last_cached_prompt`; surfaced in the diagnostics sidecar (`cached_prompt_tokens` + `cache_hit_rate`) and the context-inspector popover. Phases 2/3 (prompt reorder, `cache_control` hints) explicitly out of scope.
- **#310 — tool usage:** `tool_telemetry.py` subscriber → `workspace/tool_usage.jsonl` (source classification core/skill/mcp, outcome from result-text prefix); `make tool-usage-report` ranks tools + flags never-called core/skill tools.
- **#409 — reflection:** `ReflectionResult` now carries per-round judge tokens (were discarded); `reflection_metrics.py` subscriber → `workspace/reflection/metrics.jsonl` with outcome buckets, retry count, judge token cost, first-vs-final response delta (char + Jaccard overlap), critique fingerprint; `make reflection-stats`. An explicit `reflection_exhausted` flag is set before `_reflection_skip` nulls `last_reflection` so `loop_exhausted` survives.
- Docs: `config.md` (telemetry group), `tools.md`, `reflection.md`, `context-composer.md`, CLAUDE.md key-files.

## Test Plan
- [x] `make test` — 3048 passed (34 new: provider parsing, tool telemetry, reflection metrics + through-`run_agent_turn` bucket integration)
- [x] `make check` — clean (ruff + pyright + tsc)
- [x] `make tool-usage-report` / `make reflection-stats` — run against an empty log without error
- [ ] **Live (post-merge):** run turns in the web UI; confirm records land in the two sidecars with sane fields, the report/stats rank them, and the popover shows the Cached row rising as the prefix stabilizes.

## References
- Spec: `docs/dev-sessions/2026-07-23-0917-instrumentation-measure-first/spec.md`
- Plan: `docs/dev-sessions/2026-07-23-0917-instrumentation-measure-first/plan.md`
- Triage that motivated this: `docs/dev-sessions/2026-07-23-0948-issue-triage/triage.md` (roadmap #622)
- Closes #310
- Closes #409
- #480 — **Phase 1 only**; issue stays open for Phases 2/3 (gated on this data)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -184,6 +184,8 @@ Full doc index: [docs/index.md](docs/index.md). Hot files for navigation:
 - `prompts/` — System prompt assembly
 - `commands.py` — User-invokable commands
 - `reflection.py` — Self-reflection (Reflexion pattern)
+- `tool_telemetry.py` — Tool-usage telemetry subscriber + report (#310); `make tool-usage-report`
+- `reflection_metrics.py` — Reflection cost/effectiveness telemetry subscriber + stats (#409); `make reflection-stats`
 - `heartbeat.py`, `schedules.py`, `polling.py`
 - `notifications.py`, `notification_channels/`, `mail.py`
 - `mcp_client.py`
--- a/Makefile
+++ b/Makefile
@@ -100,6 +100,14 @@ migrate-vault-dry:
 migrate-sidecars:
 	uv run python scripts/migrate_sidecars_to_dirs.py
 
+# Ranked tool-usage report from workspace/tool_usage.jsonl (#310)
+tool-usage-report:
+	uv run python -m decafclaw.tool_telemetry
+
+# Reflection cost/effectiveness stats from workspace/reflection/metrics.jsonl (#409)
+reflection-stats:
+	uv run python -m decafclaw.reflection_metrics
+
 # Dry-run sidecar migration (show what would change)
 migrate-sidecars-dry:
 	uv run python scripts/migrate_sidecars_to_dirs.py --dry-run
--- a/docs/config.md
+++ b/docs/config.md
@@ -421,6 +421,19 @@ Skill config fields support env var overrides via the skill's `SkillConfig` meta
 
 Config CLI shows skill values as raw JSON (`config show skills`). Use `--reveal` to unmask values.
 
+### `telemetry`
+
+Instrumentation sidecars — append-only JSONL under `workspace/`, metadata only (never tool args/returns, reflection bodies, or prompt contents). Producers are fail-open EventBus subscribers. See [tools.md#tool-usage-telemetry-310](tools.md#tool-usage-telemetry-310) (#310) and [reflection.md#metrics-409](reflection.md#metrics-409) (#409).
+
+| Field | Type | Default | Env Var |
+|-------|------|---------|---------|
+| `tool_usage_enabled` | bool | `true` | `TELEMETRY_TOOL_USAGE_ENABLED` |
+| `tool_usage_path` | str | `tool_usage.jsonl` | `TELEMETRY_TOOL_USAGE_PATH` |
+| `reflection_metrics_enabled` | bool | `true` | `TELEMETRY_REFLECTION_METRICS_ENABLED` |
+| `reflection_metrics_path` | str | `reflection/metrics.jsonl` | `TELEMETRY_REFLECTION_METRICS_PATH` |
+
+Paths are workspace-relative. Enabled by default so a deployed agent starts collecting without a config edit — the intent is a week of real data. No rotation yet (append-only); retention is a follow-up. Reports: `make tool-usage-report`, `make reflection-stats`.
+
 ### `env`
 
 Arbitrary environment variables set at startup. Useful for API keys and tool-specific config that doesn't have a dedicated config field.
--- a/docs/context-composer.md
+++ b/docs/context-composer.md
@@ -173,6 +173,17 @@ The composer is mode-aware via `ComposerMode`:
 
 Each context source produces a `SourceEntry` with token estimates, item counts, and source-specific details. These are stored on `ComposerState.last_sources` and can be inspected for debugging (not published as events every turn).
 
+### Prompt-cache tokens (#480)
+
+The per-conversation diagnostics sidecar (`workspace/conversations/{conv_id}/context.json`, served by `GET /api/conversations/{id}/context`) also carries the prompt-cache figures providers report:
+
+| Field | Meaning |
+|-------|---------|
+| `cached_prompt_tokens` | Prompt tokens served from the provider's cache on the last turn |
+| `cache_hit_rate` | `cached_prompt_tokens / total_tokens_actual` (0.0 when no prompt tokens) |
+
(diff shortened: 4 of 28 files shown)


lmorchard:
Addressed all three Copilot comments:
- **context-inspector.js** — `cache_hit_rate` was truthy-gated, hiding a valid 0.0 (0% hit). Now renders `count (X%)` whenever `cached_prompt_tokens` is present, so non-cached turns show `0 (0%)` instead of a bare count.
- **tool_telemetry.py / reflection_metrics.py** — `load_records()` now streams the append-only JSONL line-by-line instead of `read_text().splitlines()`, avoiding the full-file string + list spike on these unrotated logs. (Rotation/retention remains a tracked follow-up.)

`make check` clean, 3048 tests pass.

Issue #626, opened 2026-07-23
Web terminal: add spawn→WS-attach→replay integration test


lmorchard:
Follow-up from #442 whole-branch review (non-blocking test-coverage gap).

The web-terminal pieces are each tested — the registry unit tests drive `_handle_output` directly, one real-PTY integration test covers spawn/EOF/reap, the `/ws/terminal` tests use a seeded fake session, and the `/terminal` command tests stub `spawn`. But **no single test exercises the full round trip**: a real `TerminalRegistry.spawn(...)` session served over a real `TestClient` WS attach + buffer replay, nor the smallest-viewport-wins resize end-to-end (only `_min_viewport` is unit-tested).

### Suggested
Add one integration test that spawns a real short-lived session (e.g. `/bin/echo`-style, as the existing `terminals.py` integration test does) and attaches over Starlette `TestClient` `websocket_connect`, asserting the replayed buffer + `buffer_replay_done`, and (optionally) a resize round trip.

See tests/test_terminals.py, tests/web/test_terminal_ws.py.

Issue #625, opened 2026-07-23
Web terminal: agent can orphan a PTY / create dead terminal tab via canvas tools


lmorchard:
Follow-up from #442 whole-branch review (non-blocking).

Two bounded, no-access-escalation gaps in how the agent-facing canvas tools interact with human-only terminal tabs:

1. **Orphaned PTY.** `tools/canvas_tools.py` `canvas_close_tab` calls `canvas.close_tab(...)` with no `registry`, and `canvas_clear` → `clear_canvas` removes tabs with no kill path. So the agent can remove a human-opened terminal tab (or clear the canvas) leaving the shell running until conversation-delete/server-shutdown. Not an access escalation (the agent still cannot read/write/attach the PTY), but orphaned shells count against `max_sessions_per_conv` (default 8), so repeated human-open + agent-close could eventually block new terminals with "Max sessions reached."

2. **Dead terminal tab.** The terminal `widget.json` declares `modes: ["canvas"]`, so `canvas_new_tab(widget_type="terminal", ...)` succeeds for the agent. No PTY is spawned (only the server-side `/terminal` command spawns), so the client just shows a `[session ended]` banner — cosmetic litter, no shell.

### Fix ideas
- Thread the `TerminalRegistry` into the `canvas_close_tab` / `canvas_clear` tool call sites so dropping a terminal tab kills its PTY; or have canvas mutations that drop terminal tabs emit a kill.
- Reject `widget_type == "terminal"` in the agent-facing `canvas_new_tab` (or add an `agent-createable: false` descriptor flag).

See docs/web-terminal.md and docs/dev-sessions/2026-05-06-2042-web-terminal-canvas/.

---

## Design decisions (resolved at intake, 2026-07-26/27)

Triage left three architecture questions open, which is *why* this issue was `needs-review`. All
three are now decided; the criteria below depend on them and are unreadable without them.

**D1 — the kill capability reaches agent tools as a narrowed façade on `Context`, not the registry.**
`Context` gains `terminal_registry: Any = None  # set by ConversationManager`, mirroring the existing
`request_confirmation` pattern (declared on `Context`, assigned in `conversation_manager.py` near
line 1471, sourced from `app.state.terminal_registry` — which already sits beside `state.manager` at
`http_server.py:1800`). The object assigned is **not** the `TerminalRegistry`: it is a small façade
forwarding only `.get(conv_id, tab_id)` and `.kill(session)`.

[...]

PR #627, opened 2026-07-23
feat(web): human-only PTY terminal as a canvas tab (#442)

### LINKED ISSUE #442: Web UI: shell terminal as canvas tab (xterm.js + PTY)
## Summary

Add a websocket-based shell terminal to the DecafClaw web UI, rendered as a canvas tab using xterm.js. Each terminal is a live PTY session attached over a dedicated WebSocket. Terminals are **per-conversation**, **persistent across browser reloads**, and explicitly **human-only** — the agent has no tool definitions for spawning, attaching to, or reading from terminal sessions. Multiple terminals can coexist in the same conversation's canvas; multiple browser views can attach to the same session.

The feature also establishes a generalizable **side-effect slash-command** pattern (`/terminal` is the first instance) — slash commands that produce immediate UI/server side-effects without going through the LLM or the conversation archive.

## Goals

- A user with a valid web auth token can open a fully interactive shell in their browser by typing `/terminal` in the chat input.
- The shell is a real PTY running the user's `$SHELL` (fallback `/bin/sh`) in their workspace path (or a CWD they specify).
- Browser reload restores the terminal — process keeps running, scrollback is replayed.
- The same terminal can be popped into a standalone window (`/canvas/{conv}/{tab}`) and viewed alongside the in-canvas view; both attach to the same PTY (multi-attach broadcast).
- Multiple terminals per conversation work without interference.
- Agent has zero ability to spawn, attach to, or observe terminals.

## Non-goals (deferred to future ideas)

- Cross-conversation / user-global terminals.
- Detached-then-reattachable sessions (tmux-style).
- Disk-persisted scrollback that survives server restart.
- Mobile onscreen-key helpers (Esc/Ctrl/arrows).
- Read-only agent access to the shell.
- Recording / replay.

## Design decisions

| # | Decision | Choice |
|---|----------|--------|
| 1 | Session scoping | Per-conversation (lives in `{conv_id}.canvas.json` like other tabs) |
| 2 | Lifecycle | Close-tab kills the PTY; no idle TTL; server restart kills (PTY is a child process) |
| 3 | Scrollback | In-memory ring buffer (default 10 MB), replayed on attach |
| 4 | Spawn entry point | Slash command `/terminal [cwd]` in chat input; agent cannot spawn |
| 5 | Slash-command processing | Side-effect path that bypasses LLM/archive/turn — generalizable |
| 6 | Argument support | Optional CWD argument; default = workspace path; allowed-roots whitelist |
| 7 | Multi-window attach | Multi-attach broadcast (tmux-style); smallest-viewport-wins for resize |
| 8 | Mobile | Render but no onscreen-key helpers |
| 9 | Architecture | Terminal-as-widget + dedicated WS endpoint + stdlib `pty.fork()` |

## Architecture

Three components, talking via established channels.

### Server: `decafclaw/terminals.py` (new)

In-memory `TerminalSession` registry keyed by `(conv_id, tab_id)`. Each session owns:

\`\`\`python
@dataclass
class TerminalSession:
    conv_id: str
    tab_id: str
    session_id: str           # uuid4, also embedded in widget data
    pid: int
    fd: int                   # PTY master fd
    cwd: str                  # resolved CWD at spawn
    buffer: bytearray         # ring buffer, capped at config.terminal.buffer_bytes
    attached: set[WebSocket]  # currently attached views
    viewports: dict[WebSocket, tuple[int, int]]  # per-connection cols/rows
    reader_task: asyncio.Task
    exit_status: int | None = None
\`\`\`

**Spawning** uses `pty.fork()` (in `asyncio.to_thread`). Child branch: `os.chdir(cwd)`, `os.execvpe($SHELL, [$SHELL], env)`. Parent branch: register an asyncio reader on the PTY fd via `loop.add_reader(fd, _on_pty_readable)`. Inherits the server's environment unmodified, with one override: `TERM=xterm-256color`. Initial PTY size is the kernel default (typically 80×24); corrected via `TIOCSWINSZ` when the first WS connection sends its initial `resize`.

**Reader task** reads up to 64 KB per wakeup, appends to ring buffer (drop-from-front when over cap), and broadcasts `await ws.send_bytes(chunk)` to each attached WS in `asyncio.gather(..., return_exceptions=True)`. On EOF: `os.waitpid(pid, 0)` to reap, set `exit_status`, broadcast `{type: \"session_ended\", exit_status}`, remove from registry, close all attached WS.

**Multi-attach + smallest-viewport-wins resize:** every `resize` message updates `session.viewports[ws]`. Server computes `(min_cols, min_rows)` across all viewports, calls `fcntl.ioctl(fd, termios.TIOCSWINSZ, ...)`, and broadcasts `{type: \"size_changed\", cols, rows}` so each client can letterbox if its viewport is larger.

**Replay on attach:** new WS gets the current `buffer` as binary frames followed by a `{type: \"buffer_replay_done\"}` text frame. xterm.js handles ANSI escapes in the replay correctly.

**Lifecycle:**

- Close-tab via existing `closeTabFromUi` → `close_tab` REST handler → if `widget_type == \"terminal\"`, computes confirmation message based on `attached_count` and foreground-process detection (`os.tcgetpgrp(fd) != session.pid`):
  - Idle, single view: silent close.
  - Idle, multiple views: \"Tab is open in {N} window(s). Close session?\"
  - Foreground process, single view: \"Foreground process running. Close session?\"
  - Foreground process, multiple views: combined message.
- On approval: `os.killpg(pgid, SIGHUP)`, `SIGKILL` after grace period; remove from registry.
- Server shutdown hook walks registry, SIGHUPs each session.
- Conv archive/delete: hook in `conversation_folders.py` kills any sessions for that conv.

### Server: WebSocket route `GET /ws/terminal/{conv_id}/{tab_id}`

Registered alongside `/ws/chat`. Validates session cookie via `get_current_user`; rejects with 4001 if missing. If session not found, sends `session_ended` and closes. Otherwise: register attachment → replay buffer → bidirectional pump.

**Protocol:**

| Direction | Type | Shape |
|-----------|------|-------|
| C→S | `input` | `{\"type\": \"input\", \"data\": \"<utf-8>\"}` |
| C→S | `resize` | `{\"type\": \"resize\", \"cols\": N, \"rows\": M}` |
| C→S | `ping` | `{\"type\": \"ping\"}` |
| S→C | _binary_ | Raw stdout/stderr bytes |
| S→C | `session_ended` | `{\"type\": \"session_ended\", \"exit_status\": N}` |
| S→C | `size_changed` | `{\"type\": \"size_changed\", \"cols\": N, \"rows\": M}` |
| S→C | `buffer_replay_done` | `{\"type\": \"buffer_replay_done\"}` |

### Server: side-effect command endpoint `POST /api/commands/{name}`

Single mounted route, looks up `name` in registry, validates auth, builds `CommandContext`, dispatches. Returns `{ok, ...result}` or `{ok: false, error}` (HTTP 200 in both cases for clean toast handling).

### Client: terminal widget at `web/static/widgets/terminal/`

`widget.json`:

\`\`\`json
{
  \"name\": \"terminal\",
  \"description\": \"Live PTY shell session attached over WebSocket. Created by /terminal slash command — agent cannot spawn.\",
  \"modes\": [\"canvas\"],
  \"accepts_input\": false,
  \"data_schema\": {
    \"type\": \"object\",
    \"required\": [\"session_id\"],
    \"properties\": {
      \"session_id\": {\"type\": \"string\", \"minLength\": 1},
      \"cwd\":        {\"type\": \"string\"},
      \"shell\":      {\"type\": \"string\"}
    },
    \"additionalProperties\": false
  }
}
\`\`\`

`widget.js` — `<dc-widget-terminal>` Lit component:

- `connectedCallback`: mount xterm.js, apply fit-addon, open WS to `/ws/terminal/{conv}/{tab}`. `ResizeObserver` on container → fit-addon recomputes → send `resize`.
- WS `onmessage`: binary → `term.write(data)`; text → JSON control.
- xterm.js `onData`: send `input`.
- `disconnectedCallback`: close WS, dispose xterm.
- `session_ended`: render `[session ended · exit N]` banner; stop accepting input.

Auto-reconnect with exponential backoff on unexpected WS close (1s/2s/4s/.../30s, max 5 attempts). On reconnect, server replays buffer; lost input is lost (no retry queue).

Theming: xterm.js theme populated from CSS custom props (`--pico-color`, `--pico-card-background-color`); subscribes to `theme-toggle.js` events for live updates.

### Side-effect command pattern

\`\`\`python
@side_effect_command(\"terminal\")
async def cmd_terminal(ctx: CommandContext, args: dict) -> dict:
    cwd = args.get(\"cwd\") or str(ctx.config.workspace_path)
    _validate_cwd_against_allowed_roots(ctx.config, cwd)
    session = await terminals.spawn(ctx, conv_id=args[\"conv_id\"], cwd=cwd)
    result = await canvas.new_tab(
        ctx.config, args[\"conv_id\"],
        widget_type=\"terminal\",
        data={\"session_id\": session.session_id, \"cwd\": cwd},
        emit=ctx.emit_canvas,
    )
    return {\"tab_id\": result.tab_id, \"session_id\": session.session_id}
\`\`\`

**Constraints baked in:**

- Side-effect commands **never** trigger an agent turn, **never** write to archive, **never** invoke the LLM.
- Run with caller's web-auth privileges; **no access to agent's tool registry** — no escalation through \"the bash tool internally.\"
- `CommandContext` is a distinct type from agent `Context`.

**Client:** `chat-input.js` has a `SIDE_EFFECT_COMMANDS` table mapping `/name` → `{endpoint, parseArgs}`. On submit, intercepts matching commands, POSTs, never touches the message stream.

**Distinction from existing user-invocable skill commands** (e.g. `/dream`):

| | Skill command | Side-effect command |
|--|---------------|---------------------|
| Triggers a turn? | Yes | No |
| LLM involved? | Yes | No |
| Written to archive? | Yes | No |
| Use case | \"Have the agent do something\" | \"Have the UI/server do something deterministically\" |

## Configuration

\`\`\`python
@dataclass
class TerminalConfig:
    enabled: bool = True
    buffer_bytes: int = 10 * 1024 * 1024
    default_cwd: str | None = None
    allowed_cwd_roots: list[str] = field(default_factory=list)  # empty → [workspace, \$HOME]
    shell_override: str | None = None
    max_sessions_per_conv: int = 8
\`\`\`

Resolution: dataclass defaults → `data/{agent_id}/config.json` → env (`TERMINAL_ENABLED`, etc.). `enabled=false` makes `/terminal` return \"feature disabled\" toast and the WS route 404.

## Security

- `enabled=true` is the trust boundary. Web auth cookie is the gate; this is human-only by design.
- Agent has **zero tool definitions** for any terminal operation. `decafclaw/terminals.py` is not imported anywhere in `decafclaw/tools/` or `decafclaw/skills/`.
- WS endpoint validates session cookie via `get_current_user` (same path as `/ws/chat`). No token-in-URL fallback.
- CWD validation: `Path(requested_cwd).resolve()` must be within one of `allowed_cwd_roots` (default `[workspace_path, \$HOME]`). Defense-in-depth, not a security guarantee.
- Per-conv session cap counts only live sessions (steady-state ceiling); `/api/commands/terminal` returns `{ok: false, error: \"max sessions reached\"}` when hit.
- Logging at INFO: `username, conv_id, tab_id, cwd, shell, pid` on spawn; exit status on session end. Buffer contents never logged.

## Testing

`tests/test_terminals.py` (new) and `tests/web/test_side_effect_commands.py` (new):

- **Unit, no PTY:** `TerminalSession` accepts dependency-injected `pty_module`/`os_module`. Tests stub `pty.fork()` and `os.read()` to drive the reader task with canned bytes.
- **Integration, real PTY:** one test that spawns `/bin/echo hello` (not `\$SHELL` — too slow and shell-rc-noisy), reads output, asserts cleanup on EOF.
- **WS protocol:** Starlette `TestClient` + `WebSocketTestSession`; auth rejection, session-not-found, replay, input/resize.
- **Side-effect command:** registry, auth gate, allowed-roots, \"no archive write\" invariant.
- **No-agent-turn invariant:** regression test asserts `/api/commands/terminal` does not invoke `ConversationManager.enqueue_turn`.
- **Test speed:** patch `terminals.spawn` in any test using `make_app()` so fixtures don't accidentally spawn real PTYs.

## File-level change summary

**New:**
- `src/decafclaw/terminals.py` — `TerminalSession`, registry, spawn/attach/detach/kill, reader task.
- `src/decafclaw/web/side_effect_commands.py` — registry decorator, `CommandContext`, `POST /api/commands/{name}` route.
- `src/decafclaw/web/static/widgets/terminal/widget.json` + `widget.js`.
- `tests/test_terminals.py` + `tests/web/test_side_effect_commands.py`.
- `docs/web-terminal.md` — feature reference.
- `docs/side-effect-commands.md` — pattern doc.

**Modified:**
- `src/decafclaw/config_types.py` — `TerminalConfig` dataclass.
- `src/decafclaw/config.py` — env var resolution.
- `src/decafclaw/runner.py` — register routes; shutdown hook.
- `src/decafclaw/web/websocket.py` — register WS route alongside chat.
- `src/decafclaw/web/conversation_folders.py` — kill sessions on conv archive/delete.
- `src/decafclaw/web/static/components/chat-input.js` — `SIDE_EFFECT_COMMANDS` interception.
- `src/decafclaw/web/static/components/canvas-panel.js` — verify widget-host doesn't tear down on tab switch.
- `src/decafclaw/canvas.py` — close-tab path: detect `widget_type == \"terminal\"` and route through terminal kill flow.
- `package.json` (vendor build) — add `xterm`, `@xterm/addon-fit`, `@xterm/addon-serialize`, `@xterm/addon-web-links`.
- `docs/web-ui.md`, `docs/index.md`, `CLAUDE.md`.

## Future ideas (deferred)

1. **Read-only agent access to the shell.**
2. **Disk-persisted scrollback** (survives server restart).
3. **Detached-terminals model** (close-tab-detaches; tmux-style).
4. **Cross-conversation user-global terminals.**
5. **Onscreen key helpers for mobile.**
6. **Vendor bundle splitting / per-widget lazy-loading.**
7. **Standalone-window-only terminals** (skip canvas tab entirely).
8. **Recording / replay** (\`script(1)\`-style).

## Open questions

- **Widget-host re-mount on tab switch:** unknown until implementation. If `<dc-widget-host>` re-mounts widgets on `activeTabId` change, the terminal would thrash WS connections on every tab click. Mitigation: preserve widget instances across switches, OR detect \"hidden but still in DOM\" in the widget. Resolve during implementation.
- **Standalone-window widget loading:** verify `/canvas/{conv}/{tab}` standalone page loads the same widget bundle as the embedded canvas (almost certainly does — `iframe_sandbox` works in both).
- **PTY behavior on macOS vs Linux:** primary dev = macOS; deploy = Linux (`decafclaw@lmorchard`). Test on both.
- **Multi-attach input interleaving:** typing in a background window corrupts the foreground session's command line. Acceptable for power-user use; flag in `docs/web-terminal.md`.

lmorchard:
Closes #442.

Adds a WebSocket-backed PTY shell to the web UI, rendered as a canvas tab via xterm.js, opened with `/terminal [cwd]`. **Human-only by design** — the agent has no tool, no code path, and no `ctx` handle to spawn/attach/read a terminal (enforced by an import-boundary test).

## What's here
- **`terminals.py`** — in-memory `TerminalRegistry` of PTY sessions (ring buffer, multi-attach broadcast, smallest-viewport-wins resize, EOF reap).
- **`/ws/terminal/{conv}/{tab}`** — cookie-authed WebSocket; replays scrollback then streams; guarded control-frame parsing.
- **`<dc-widget-terminal>`** — xterm.js client: connect→replay→attached state machine, capped-backoff reconnect, theme-at-mount, session-ended banner.
- **`/terminal` command** — handled **server-side** in `_handle_send`: spawns the PTY + creates the canvas tab with **no LLM turn and no archive write** (invariant is test-asserted).
- **Canvas keep-alive host** — each tab keeps its own mounted `<dc-widget-host>` (per-tab, keyed), so a terminal's socket survives tab switches and two same-type tabs no longer alias onto one instance.
- **Lifecycle kills** — close-tab (client confirm → PTY kill), server shutdown, conversation delete.
- **`TerminalConfig`** (`TERMINAL_*`) — `enabled` (trust boundary), `buffer_bytes`, `default_cwd`, `allowed_cwd_roots`, `shell_override`, `max_sessions_per_conv`.
- Docs: `docs/web-terminal.md` + web-ui/index/CLAUDE updates.

## Notable design decision
Spawning uses **`os.posix_spawn(..., setsid=True)` + `os.openpty()`**, not `pty.fork()` — forking a multi-threaded interpreter emits an intrinsic CPython deadlock-risk `DeprecationWarning`, and our conventions forbid suppressing warnings. posix_spawn sidesteps the fork entirely; CWD is set via a `sh -c 'cd … && exec <shell>'` trampoline. (Verified on macOS; Linux supports `setsid` via glibc.) See `docs/web-terminal.md` and the session notes.

## Testing
Full suite green (3076 passing); `make check`/`check-js` clean. Server-side pieces are unit-tested (registry logic incl. a real-PTY integration test, WS route via TestClient, command invariants incl. the no-turn/no-archive assertions). JS has no unit harness — the widget + keep-alive host were `check-js` + logic-reviewed.

**Pending before merge:** a live browser smoke test (two terminals, tab-switch persistence, cross-tab isolation, close-kill) — the one validation unit tests can't cover.

## Process
Built via subagent-driven development (10 tasks, per-task spec+quality review, whole-branch review). Dev-session docs under `docs/dev-sessions/2026-05-06-2042-web-terminal-canvas/`.

## Follow-ups (non-blocking, filed from the whole-branch review)
- #625 — agent can orphan a PTY / create a dead terminal tab via canvas tools (bounded, no access escalation)
- #626 — add a spawn→WS-attach→replay integration test (the one untested seam)


lmorchard:
## Live browser smoke test — done ✅ (found + fixed 2 bugs)

Drove the real web UI (Playwright) against an isolated server (temp `DATA_HOME`, heartbeat off, MM off). End-to-end validated: `/terminal` spawns a real `zsh` PTY (posix_spawn+setsid) + canvas tab → widget mounts → WS attaches + replays → **live prompt** → **interactive I/O** (`echo` → output) → **two terminals fully isolated** (own sessions, zero content bleed — the aliasing fix + keep-alive; hidden tab stayed attached) → **close-tab confirm + PTY kill** (closed tab's shell died, the other survived).

The smoke test caught two bugs that unit tests, `check-js`, and all three review layers missed:

1. **🔴 Blocker** (`f3a52cf`) — the vendored xterm addon bundles exposed classes only under `default` (esbuild CJS interop), so the widget's `import { FitAddon }` threw at load and **every terminal rendered "unavailable."** `check-js` missed it because tsc trusts the package's named-export types. Fixed with normalize entry files re-exporting named + default (mirroring `leaflet-entry.js`).
2. **🟠 UX** (`692efd7`) — closing a terminal tab popped **two** confirm dialogs (canvas-panel's generic prompt + Task 9's terminal-specific one). Consolidated to a single terminal-aware confirm in `closeTabFromUi`, covering every close path.

Both re-verified in-browser after the fix. Suite still green (3076 passing), `make check`/`check-js` clean.

Takeaway worth noting for future widget work: JS runtime/bundle-interop failures are invisible to tsc + Python unit tests + diff review — a real-browser smoke is the only net (see #626 for a JS/integration-test follow-up).

PR #629, opened 2026-07-23
fix(tool-search): rank name matches above description matches (#526)

### LINKED ISSUE #526: tool_search keyword scoring: 'wait' matches heartbeat_trigger before wait
Surfaced during the 2026-05-16 evals-renovation session writing tool-deferral evals.

## Problem

\`tool_search(query=\"wait\")\` returns \`heartbeat_trigger\` as the top match, not the tool actually named \`wait\`. The query token \"wait\" appears in \`heartbeat_trigger\`'s description (\"without waiting for the next scheduled tick\") and presumably gets a higher score than the bare name match against \`wait\`.

## Expected behavior

Name matches (exact or near-exact) should outrank description matches. \`tool_search(query=\"wait\")\` should return \`wait\` first.

## Repro

\`\`\`
uv run python -c \"
import asyncio
from decafclaw.config import load_config
from decafclaw.context import Context
from decafclaw.events import EventBus
from decafclaw.tools.search_tools import tool_tool_search
async def main():
    config = load_config()
    ctx = Context(config=config, event_bus=EventBus())
    print(await tool_tool_search(ctx, query='wait'))
asyncio.run(main())
\"
\`\`\`

(Approximation — actual repro may need a different invocation depending on how \`tool_search\` is wired in the runtime vs. a CLI.)

## Why it matters

This is a stealth bias in the deferred-tool surface. Agents asking for a tool by exact name will be steered toward tools with that word elsewhere in their description. The audit's #344 (tool-deferral evals) hits this directly — eval prompts that name a tool can route to the wrong tool.

## Suggested fix

Boost the score for name-token matches relative to description-token matches. Possibly: \`score = 2 * name_matches + description_matches\` instead of an undifferentiated union.

## Priority / Size

P2 / S.

lmorchard:
## Problem

`tool_search` returned keyword matches in deferred-pool order with no scoring. A query naming a tool (e.g. `wait`) could be outranked by an unrelated tool that merely mentions the word in its description — the issue's example: `tool_search(query="wait")` surfaced `heartbeat_trigger` (whose description says "without *waiting* for the next scheduled tick") ahead of the tool actually named `wait`.

This is a stealth bias in the deferred-tool surface: agents asking for a tool by exact name get steered toward tools with that word elsewhere in their description.

## Fix

Score each keyword match instead of returning an undifferentiated union:

- exact name match → **3**
- partial name match → **2**
- description hit → **+1** (summed with any name score)

Skills and tools are each rendered in ranked (score-descending) order. `max_results` truncation now keeps the highest-scored matches. Ties preserve discovery / pool order (stable sort). `select:` exact lookup is unchanged.

Slightly richer than the issue's suggested `2*name + desc` — the exact-name tier is a stronger, correct signal and stays simple.

## Verification

- **Test-first**: added `TestKeywordScoring` (name-outranks-description + truncation-keeps-highest-scored). Both failed before the fix, pass after.
- Reproduced the issue scenario against the real `wait` / `heartbeat_trigger` definitions — `wait` now ranks first.
- `make lint typecheck` clean; 108 tests pass across the tool-related suites.
- Updated `docs/tool-search.md` with the ranking rules (same-PR doc convention).

No eval added: this is deterministic scoring fully covered by unit tests — no LLM routing decision changed.

Closes #526


lmorchard:
Thanks — addressed all four in 4c9246e:

- **Exact-vs-partial tie (comments 1, 3, 4):** real bug. Scoring is now a `(name_tier, description_hit)` tuple compared lexicographically, so an exact-name match strictly outranks a partial-name match regardless of description hits. The old additive `3 / 2 / +1` let partial-name + description (`2+1`) tie exact-name-only (`3`). Added a test with `wait_for` (partial name + "wait" in description) placed *earlier* in the pool than exact `wait`, so a stable sort can't mask a regression.
- **O(pool × skills) scan (comment 2):** owning-skill descriptions now come from a dict built once before the pool walk.
- Doc updated to describe the strict tiering (name tier dominates; description hit only breaks ties within a tier).

lmorchard:
Addressed the second-pass comment in 4d05d9c: the `max_results` budget is now applied across skills and tools **together**, ranked by score, so a low-scoring (e.g. description-only) skill match can no longer evict a higher-scoring tool match. Skills still win ties (lighter pointers), and rendering still groups skills then tools. Added a test (`test_budget_keeps_high_scored_tool_over_low_scored_skill`) and updated the doc claim to match.

PR #630, opened 2026-07-23
fix(tools): never suggest the unknown tool name back in did-you-mean hint (#355)

### LINKED ISSUE #355: Tool registry: did-you-mean error suggests the same unknown tool name
Not directly part of the #240 eval-coverage split, but surfaced by the eval audit run — see [audit doc](https://github.com/lmorchard/decafclaw/blob/eval-coverage/docs/dev-sessions/2026-04-24-0941-eval-coverage/evals-audit.md) in #338.

## Problem

When the agent calls an unknown tool, the tool-dispatch layer produces an error like:

```
[error: unknown tool 'project_advance'. Did you mean: project_advance, project_task_done, project_note, project_create, project_add_steps. Use the exact name from your tool list. To discover available...]
```

The suggested alternative `project_advance` is **identical to the unknown tool name the agent tried**. That's clearly a bug in the did-you-mean logic — probably the suggestion ranking doesn't exclude the input string when it happens to match an existing tool name.

Observed during eval run `evals/results/2026-04-24-1015-default/` on the project-skill test "writes spec when asked".

## Scope

- Find the did-you-mean code path (likely in `src/decafclaw/tools/__init__.py` or `tool_registry.py`).
- Understand why `project_advance` was flagged as unknown but then returned as a suggestion. Plausible cause: name collides with some deferred/unactivated tool, or normalization folds two different strings together.
- Fix the suggestion logic to exclude any suggestion string equal (case-insensitive) to the unknown input.
- Add a unit test.

## Acceptance criteria

- The error message never suggests the exact unknown name back to the caller.
- Unit test covers the regression.

## Priority / Size

P2 / XS.

lmorchard:
## Summary

When the agent calls an unknown tool whose name is also present in the candidate pool (a deferred/unactivated tool or MCP name), `_suggest_tool_names` returned that exact name as the top suggestion — because `difflib.get_close_matches` treats an exact string as a ratio-1.0 match. The error then read:

```
[error: unknown tool 'project_advance'. Did you mean: project_advance, project_task_done, ...]
```

i.e. it suggested the exact name the agent just tried. The suffix-match loop already guarded with `cand != name` (case-sensitive), but the difflib loop had no self-exclusion.

## Fix

Filter the candidate set to drop any entry case-insensitively equal to the unknown input before matching, in `src/decafclaw/tools/__init__.py::_suggest_tool_names`. The suffix loop's now-redundant `cand != name` guard is removed in favor of the single filter.

## Testing

- `test_suggest_tool_names_excludes_exact_input` — name present in candidate pool is never suggested back (test-first: red before fix, green after).
- `test_suggest_tool_names_excludes_case_insensitive_input` — a candidate differing only in case is treated as the same name.
- Existing suggestion tests (typo close-match, MCP suffix-match) still pass.
- `make lint` ✓, `make typecheck` ✓ (0 errors), `make test` ✓ (3052 passed).

Deterministic error-message path → unit tests, no eval needed.

Closes #355

### DIFF
--- a/src/decafclaw/tools/__init__.py
+++ b/src/decafclaw/tools/__init__.py
@@ -174,8 +174,11 @@ def _suggest_tool_names(name: str, candidates: set[str], max_results: int = 5) -
     # present in the candidate pool (e.g. a deferred/unactivated tool),
     # difflib would return it as an exact ratio-1.0 match. Case-insensitive
     # so a candidate differing only in case is treated as the same name.
+    # `isinstance(c, str)` guards against a malformed tool def leaving a
+    # None in the pool (deferred_names uses `.get("name")` with no default)
+    # — a suggestion hint must never crash the unknown-tool error path.
     name_lower = name.lower()
-    candidates = {c for c in candidates if c.lower() != name_lower}
+    candidates = {c for c in candidates if isinstance(c, str) and c.lower() != name_lower}
     if not candidates:
         return []
     suggestions: list[str] = []
--- a/src/decafclaw/tools/__init__.py
+++ b/src/decafclaw/tools/__init__.py
@@ -170,13 +170,21 @@ def _suggest_tool_names(name: str, candidates: set[str], max_results: int = 5) -
     """
     if not candidates:
         return []
+    # Never suggest the unknown name back to the caller (#355): if it's
+    # present in the candidate pool (e.g. a deferred/unactivated tool),
+    # difflib would return it as an exact ratio-1.0 match. Case-insensitive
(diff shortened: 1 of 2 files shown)


lmorchard:
Addressed the Copilot review comment in 4f1e474: added an `isinstance(c, str)` guard to the candidate filter in `_suggest_tool_names` so a `None` from a malformed deferred tool def (`deferred_names` uses `.get("name")` with no default) can't crash the unknown-tool error path. Added a regression test with a `None` in the candidate pool. Full suite green (3053 passed).

PR #631, opened 2026-07-23
fix(background): close subprocess transport deterministically (#605)

### LINKED ISSUE #605: Flaky PytestUnraisableExceptionWarning: leaked asyncio subprocess transport in test suite
## Symptom
`make test` intermittently ends with `1 warning`:
\`\`\`
PytestUnraisableExceptionWarning: Exception ignored in: <function BaseSubprocessTransport.__del__>
\`\`\`
(The inner traceback is CPython 3.13 GC-teardown noise: `'PosixPath' object has no attribute '_parts_normcase_cached'`.)

pytest attributes it to whatever test happens to be running on the xdist worker when the orphaned transport is garbage-collected, so the *attributed* test name varies and is misleading.

## It's pre-existing and flaky (evidence)
Measured on the `feat-blog-ideas-skill` branch:
- Full suite **with** the new `tests/test_blog_ideas_skill.py`: 1 warning, 3/3 runs (attributed to that test).
- Full suite **excluding** that file (≈ clean main): 1 warning, **3/3 runs** — and a separate earlier run showed 0. So the warning exists independent of any one test and surfaces non-deterministically based on GC timing.

The blog-ideas skill spawns no subprocess; it's uninvolved. The real source is a test that creates an asyncio subprocess and relies on GC for cleanup — likely among the `claude_code` / `background` / `mcp_client` tests (the three modules in the tree that call `asyncio.create_subprocess_*` / spawn stdio servers).

## Why it matters
Project policy is zero-tolerance for warnings; a flaky warning erodes that signal and makes `make test` non-deterministically "dirty."

## Suggested fix
Find the test leaving the subprocess transport unclosed and close it deterministically (await `proc.wait()` / `transport.close()` in a finally or fixture teardown), or add a fixture that drains/closes the event loop's subprocess transports at teardown. A `filterwarnings = error::pytest.PytestUnraisableExceptionWarning` gate (once clean) would prevent regressions.

Size: S–M. Priority: P2.

lmorchard:
## Summary
- Fixes the flaky `PytestUnraisableExceptionWarning` that intermittently ended `make test` with `1 warning` (`Exception ignored in: <function BaseSubprocessTransport.__del__>`).
- Project policy is zero-tolerance for warnings; a *flaky* warning erodes that signal and makes `make test` non-deterministically "dirty."

## Root cause (evidence-based)
`BackgroundJobManager` spawns real asyncio subprocesses (`create_subprocess_shell`). `Process.wait()` reaps the child but never closes the subprocess **transport**. asyncio schedules its own `transport.close()` (so `is_closing()` misleadingly reads `True`), but that scheduled close races pytest-asyncio's short-lived per-test event-loop teardown. The transport survives its loop, and `BaseSubprocessTransport.__del__` runs at **interpreter shutdown** — after the loop is closed → `RuntimeError('Event loop is closed')` → unraisable → the warning, misattributed to whatever test the xdist worker was running.

Localized deterministically (`-n0 -W error::pytest.PytestUnraisableExceptionWarning`): `test_background_tools.py` warned **5/5**; all other suspects (`mcp`, `claude_code`, `background_wake`, `process_tool_media`) **0/5**. Causality proven by A/B over the full file, 20 runs each: **fix disabled → 19/20 warned; fix enabled → 0/20**.

## Changes
- **`skills/background/tools.py`** — new `_close_transport(job)` helper called in `_run_reader`'s `finally` (runs inside the live loop for both natural completion and cancellation). Closes `job.process._transport` so `_closed` is set and `__del__` is a no-op. `_transport` is asyncio-internal (no public accessor on `Process`); `close()` is idempotent; fail-open.
- **`pyproject.toml`** — `filterwarnings = ["error::pytest.PytestUnraisableExceptionWarning"]` regression gate: a future leaked resource now fails the suite deterministically instead of flakily dirtying it (as the issue recommends).
- **`tests/test_background_tools.py`** — 3 deterministic unit tests of `_close_transport` (close / fail-open on missing `_transport` / fail-open when `close()` raises). The end-to-end regression is caught by the gate (a GC-timing assertion would be flaky; `is_closing()` doesn't distinguish since asyncio calls `close()` itself).

## Test Plan
- [x] `make check` clean
- [x] `make test` — **3053 passed**, and the whole suite is clean under the new gate across 5+ runs (previously an intermittent `1 warning`)
- [x] A/B verified the fix is causal (19/20 → 0/20)

## References
- Spec/notes: `docs/dev-sessions/2026-07-23-1258-leaked-subprocess-transport/` (root-cause narrative in `notes.md`)
- Closes #605

### DIFF
--- a/docs/dev-sessions/2026-07-23-1258-leaked-subprocess-transport/notes.md
+++ b/docs/dev-sessions/2026-07-23-1258-leaked-subprocess-transport/notes.md
@@ -1,0 +1,51 @@
+# Notes — #605 leaked subprocess transport
+
+## Root cause (systematic-debugging, evidence-based)
+
+**Symptom:** `make test` intermittently ends with `1 warning` —
+`PytestUnraisableExceptionWarning: Exception ignored in: <function BaseSubprocessTransport.__del__>`.
+
+**Localization (deterministic reproduction found):**
+`uv run pytest tests/<file>.py -n0 -W error::pytest.PytestUnraisableExceptionWarning`, 5 runs each:
+
+| file | warned |
+|------|--------|
+| `test_background_tools.py` | **5/5** |
+| test_background_wake_integration / test_mcp / test_claude_code_output / test_claude_code_sessions / test_process_tool_media | 0/5 |
+
+So the leak is in `test_background_tools.py` — the only tests that spawn a **real** asyncio subprocess via `BackgroundJobManager` (`asyncio.create_subprocess_shell`). No single test warns alone (0/3 each); the file warns as a whole → classic "leaks in test X, GC'd during a later test / session teardown."
+
+**Full failing traceback** pinned the mechanism:
+- `24 passed`, warning raised at `_pytest/unraisableexception.py:95 cleanup` — i.e. **session teardown**, not any test.
+- Inner: `asyncio/base_subprocess.py:130 in __del__ → RuntimeError('Event loop is closed')`. (The issue observed `'PosixPath' object has no attribute '_parts_normcase_cached'` — same class of CPython-3.13 teardown-after-loop-closed noise.)
+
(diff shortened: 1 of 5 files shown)


lmorchard:
Good catch — fixed in the latest push. Wrapped the finalize step in its own `try/finally` so `_close_transport(job)` runs even when `_finalize_job` raises or the reader is cancelled mid-finalize:

```python
finally:
    try:
        if not cancelled:
            await _finalize_job(job)
    finally:
        _close_transport(job)
```

Added `test_transport_closed_even_when_finalize_raises` covering exactly that path (patches `_finalize_job` to raise, asserts the transport is still closed). `make check` clean; full suite `3058 passed` under the new warnings-as-error gate.

PR #633, opened 2026-07-23
fix(conversation-search): stemmed token matching + relevance ranking (#535)

### LINKED ISSUE #535: conversation_search uses substring match — fails on plurality/inflection mismatches
Surfaced during the 2026-05-16 evals-renovation session writing \`evals/conversation.yaml\`.

## Problem

\`tool_conversation_search\` (`src/decafclaw/tools/conversation_tools.py:11`) uses case-insensitive substring matching: \`query_lower in content.lower()\`. The seed message was \`\"I'm planning to switch our embedding provider from OpenAI to Vertex\"\` (singular). The agent's query was \`\"embedding providers\"\` (plural). Zero hits. Agent returned \`\"I could not find any previous mentions of embedding providers\"\` even though the conversation literally discussed exactly that.

This is the same kind of bug real users will hit: ask the agent \"what did I say about embedding providers\" → tool returns nothing → agent fabricates or denies → user is confused because they DID mention it.

## Why this matters

`conversation_search` is the only mechanism for the agent to recall historical context that's outside the active window or post-compaction. Substring-only matching makes it fragile to natural language variation. Failures here look like \"the agent forgot\" rather than \"the search query didn't match.\"

## Repro

1. Append to conversation archive: \`{\"role\":\"user\",\"content\":\"My favorite color is blue\"}\`
2. Call \`tool_conversation_search(ctx, query=\"colors I like\")\`
3. Returns \"No conversation history found\" — even though \"blue\" + \"color\" is right there.

## Proposed approaches

**Option A (cheap):** Document the substring-exact behavior in the tool description so the LLM knows to use short, exact-token queries. The model adapts. Costs nothing; no behavior change. Doesn't fix the underlying limitation.

**Option B (real fix):** Token-aware matching. Tokenize both query and content, do set-intersection scoring like `preempt_search.py` does. Returns ranked results; threshold for inclusion. Compatible with the existing substring API (treats single-token queries the same way).

**Option C (semantic):** Use embeddings. Big lift — needs an index over archive content, build/update on append. Overkill for the size of typical archives.

Recommended: B. A is a stop-gap.

## Acceptance criteria

- \`conversation_search(query=\"colors I like\")\` returns a match for a seeded \`\"My favorite color is blue\"\` message.
- Existing single-token substring queries (e.g., \`\"Mozilla\"\`) still work.
- Score-based ranking so higher-overlap entries surface first.
- Document the matching mode in the tool description.

## Priority / Size

P2 / S.

Tagged: not labelled \`evals\` because this is a product-side change, not an eval-harness issue. The renovation surfaced it but the fix lives in `conversation_tools.py`.

lmorchard:
## Problem

`conversation_search` used case-insensitive **substring** matching (`query_lower in content.lower()`). A query whose wording differed from the archived text by inflection returned zero hits — the issue's headline case: `"embedding providers"` (plural) missed a message that said `"embedding provider"` (singular), and the agent would deny ever discussing it. This tool is the agent's only recall path for history outside the active window / post-compaction, so failures read as "the agent forgot."

## Fix (issue's Option B + zero-regression substring)

A message matches if **either** branch fires, and the score ranks results:

- **Substring branch** (dominant score `1000`): preserves 100% of prior behavior, including mid-word matches like `config` → `configuration`.
- **Stemmed token-overlap branch** (score = overlap count): tokenize via the existing `preempt_search.tokenize` (stopwords, ≥3 chars), then stem with snowballstemmer (Porter2). Bridges plural/inflected queries (`colors`~`color`, `providers`~`provider`).
- Results ranked by score, so higher-overlap entries surface first; ties keep archive order.

Stemming is applied **only** in `conversation_search` — preempt-search tool-promotion behavior is untouched (it has its own evals).

## Verification

- **Test-first**: added the two issue repros (`"colors I like"` → `"My favorite color is blue"`; plural query → singular seed) plus regression guards (mid-word substring, whole-word `pelican`/`osprey`, relevance ranking, no-overlap → no match). All green.
- Full suite: **3066 passed**. `make lint typecheck` clean.
- Added an inflection-tolerance case to `evals/conversation.yaml` and corrected the case comment + `docs/` that mislabeled this tool as "semantic" (it never was).

Dependency: `snowballstemmer>=2.2.0` (pure-Python, MIT, no transitive deps).

Note: the eval case is added but not run here (needs live model calls); happy to run `make eval FILE=conversation` on request.

Closes #535

### DIFF
--- a/docs/conversations.md
+++ b/docs/conversations.md
@@ -99,7 +99,7 @@ Place a `COMPACTION.md` file at `data/{agent_id}/COMPACTION.md` to customize the
 ### Tools
 
 - **`conversation_compact`** — manually trigger compaction without waiting for the token budget to be exceeded
-- **`conversation_search`** — search past conversations using semantic search (across all archived conversations, not just the current one)
+- **`conversation_search`** — search past conversations by stemmed word overlap plus substring, ranked by relevance (across all archived conversations, not just the current one). Stemming means plural/inflected queries match — `"embedding providers"` finds `"embedding provider"`.
 
 ## Web UI conversations
 
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -51,7 +51,7 @@ Always-activated skill for the unified knowledge base. See [Vault](vault.md).
 
 | Tool | What it does |
 |------|--------------|
-| `conversation_search` | Search past conversation archives (semantic) |
+| `conversation_search` | Search past conversation archives (stemmed word overlap + substring) |
 | `conversation_compact` | Manually trigger conversation compaction |
 
 ## Checklist (`tools/checklist_tools.py`)
--- a/evals/conversation.yaml
+++ b/evals/conversation.yaml
@@ -10,10 +10,9 @@
 
 - name: "reaches for conversation_search on explicit history query"
   setup:
-    # `conversation_search` does case-insensitive *substring* match, so seed
-    # phrasing must contain whichever exact tokens the agent's query is
-    # likely to use. Using plural "embedding providers" + naming all three
-    # provider tokens verbatim makes the search deterministic.
+    # `conversation_search` matches by stemmed word overlap plus substring
+    # (see #535), so exact-token phrasing is no longer required — but naming
+    # the distinctive provider tokens verbatim keeps this case deterministic.
     conversation_history:
       - role: user
         content: "I'm planning to switch our embedding providers from OpenAI to Vertex."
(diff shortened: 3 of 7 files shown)


lmorchard:
Addressed in d75b5a8. One clarification on the tradeoff: correct global relevance ranking (an acceptance criterion) inherently requires scoring every message — a top-scored match can sit anywhere in the archive, so early-exit-after-10 would reintroduce the arbitrary, unranked behavior this PR is fixing. So the full scan stays.

What I *did* fix is the memory concern: result selection now uses a bounded min-heap keyed by `(score, -order)`, so memory stays O(_MAX_RESULTS) regardless of how many messages match (no more collecting the full match list). Added a test asserting the 10-result cap and the earliest-wins tie behavior.

(Worth noting for the archive-growth angle: unbounded archive size is a known gap tracked separately in CLAUDE.md — an index/streaming approach there would be the real lever if archives ever get large enough for the per-message scan to matter.)

PR #636, opened 2026-07-23
feat(context): trim preempt-skill-hint to top-scoring matches (#604)

### LINKED ISSUE #604: Reduce preempt-skill-hint noise: only surface skills within N points of the top score
Spun out of the `blog-ideas` skill work (#feat-blog-ideas-skill), where a routing experiment touched this and we reverted it to keep that PR focused. Filing the idea on its own merits.

## Observation
`ContextComposer` builds a `<preempt_skill_hint>` listing skills whose preemptive keyword match scored above threshold. On generic tokens this can list ~10 skills (e.g. a clearly-relevant skill at score 5 buried alongside several at score 1). The model doesn't distinguish scores, so a dense hint dilutes the routing signal and the model tends to default to always-loaded general tools instead of activating the relevant skill.

## Idea
In the **visible hint only**, include just the skills within a small gap of the top scorer (e.g. `score >= max(2, top_score - 1)`), always showing at least one. Low-scoring matches still land in `ctx.skills.preempt_matches` for `classify_tools` promotion — only the hint text is trimmed.

Reference implementation (reverted from the blog-ideas branch) added, after `ctx.skills.preempt_matches = matched_names` in `context_composer.py`:

\`\`\`python
top_score = scored[0]["score"] if scored else 0
hint_top = [e for e in top if e["score"] >= max(2, top_score - 1)]
if not hint_top:
    hint_top = top[:1]  # always show at least one
hint_names = {e["name"] for e in hint_top}
safe_names = sorted(html.escape(name, quote=False) for name in hint_names)
\`\`\`

## Why separate
This changes routing-hint behavior for **every** skill, so it deserves its own evaluation rather than riding in on an unrelated feature. Needs: a tool_choice / routing eval demonstrating the before/after, and a check that it doesn't suppress a legitimately-relevant lower-scoring skill.

Size: S. Priority: P2.

lmorchard:
## Summary
- On generic tokens the `<preempt_skill_hint>` could list ~10 skills — a clearly-relevant skill buried among several score-1 matches. The model doesn't distinguish scores, so a dense hint dilutes the routing signal and it tends to fall back to always-loaded general tools instead of activating the relevant skill.
- This trims the **visible hint** to skills scoring within a small gap of the top match, so the strong signal isn't drowned out.

## Design decisions
- **Threshold:** `score >= max(_HINT_MIN_SCORE, top_score - _HINT_SCORE_GAP)` (named constants in `context_composer.py`, currently floor **2** / gap **1**), always keeping at least one. Hardcoded-but-named for now — trivially promotable to `config.agent.preemptive_search` later if it needs per-deployment tuning.
- **Trim is hint-only.** The full match set still lands on `ctx.skills.preempt_matches`, so a lower-scoring but genuinely-relevant skill's tools are still promoted for the turn (via `classify_tools`) — it just isn't *named* in the hint. This directly addresses the issue's "don't suppress a legitimately-relevant lower-scoring skill" concern: suppression would be a routing regression, but the tools stay reachable.
- **Diagnostics** now distinguish promoted vs hinted (`details.matches` = full promoted set, `details.hinted` = named subset, plus `promoted_count`/`hinted_count`); the log line marks which matches made the hint.

## Test Plan
- [x] `make check` clean; `make test` — 3099 passed
- [x] **Unit (deterministic):** trim-keeps-promotion (gap boundary + exclusion + promotion preserved), fallback-shows-one, tied-top-scorers; updated the existing bullet-rendering test to two equal-score skills
- [x] **Eval (real LLM):** added a routing case to `evals/tool-deferral.yaml` and ran it — `4/4 passed`, incl. "preempt skill hint activates the clearly-relevant skill" (model called `activate_skill` under the trimmed hint)

## References
- Closes #604
- Note: current `main` (post-#627/#635 terminal work) emits 2 unrelated `forkpty()` DeprecationWarnings from `pty.py` — pre-existing, not from this PR; flagged for a separate follow-up.

### DIFF
--- a/docs/preemptive-tool-search.md
+++ b/docs/preemptive-tool-search.md
@@ -37,7 +37,9 @@ Unlike the tool match, this is purely a **hint** — no auto-activation. The age
 
 The wording is intentionally bulleted (one skill per line) rather than comma-joined inline — small parent models attend to visual structure more reliably than to running prose, and the activation directive sits at the top where the model is most likely to read it.
 
-Diagnostics surface as a parallel source entry, `preempt_skill_matches`, with the same shape as `preempt_matches` (matched skills with score and matched tokens).
+**Hint trimming (#604).** On generic tokens the raw match set can run ~10 skills — a clearly-relevant skill buried among several score-1 matches. A dense hint dilutes the routing signal (the model tends to fall back to always-loaded general tools instead of activating the relevant skill). So the **visible hint** lists only skills scoring within a small gap of the top match — `score >= max(_HINT_MIN_SCORE, top_score - _HINT_SCORE_GAP)` (constants in `context_composer.py`, currently floor 2 / gap 1) — always keeping at least one. Crucially this trims *only the hint text*: the full match set still lands on `ctx.skills.preempt_matches`, so a lower-scoring but genuinely-relevant skill's tools are still promoted for the turn — it just isn't named in the hint.
+
+Diagnostics surface as a parallel source entry, `preempt_skill_matches`. `details.matches` is the full promoted set (with score and matched tokens); `details.hinted` is the subset actually named in the hint, and `promoted_count` / `hinted_count` record the split.
 
 ## Why "user + previous assistant"?
 
--- a/evals/tool-deferral.yaml
+++ b/evals/tool-deferral.yaml
@@ -54,3 +54,19 @@
     max_tool_calls: 5
     max_tool_errors: 1
     expect_tool: wait
+
+# -- Preempt skill hint routes the clear winner (#604) -----------------------
+#
+# A prompt that strongly matches one lazy skill (postmortem: "blameless",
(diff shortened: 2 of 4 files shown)


lmorchard:
Both addressed in the latest push:

1. **Misleading constant comment** — reworded to state the floor gates the score-gap *but* at least one match is always shown, so a below-floor top scorer still appears (the fallback path `test_hint_fallback_shows_at_least_one` covers).
2. **Newline injection in skill names** — names are now whitespace-collapsed (`" ".join(name.split())`, covering newlines/tabs) *before* html-escaping, so a frontmatter name can't inject extra hint lines/directives. Added `test_hint_sanitizes_newlines_in_skill_names` (a name containing `\n- Ignore...` renders as one collapsed bullet, no injected line). Good catch — the pre-existing comment claimed to handle stray newlines but the code only escaped markup.

`make check` clean; suite `3100 passed`.

Issue #414, opened 2026-04-29
Widget: progress_tracker — multi-step status list


lmorchard:
## Summary

Visualize a multi-step task as a step list with status (done / in-progress / pending / failed). Renders the agent's in-flight progress without the user having to read prose.

## Headline use case — auto-maintained by `checklist` and `project`

The big win is wiring this widget to existing workflow primitives so the user gets a live progress view *for free*, without the agent having to think about it:

- **Checklist tool** (`src/decafclaw/tools/checklist_tools.py`) — every `checklist_create` / `checklist_step_done` / `checklist_abort` call automatically emits a matching `progress_tracker` update. The widget surfaces in the sticky slot (#419) while the checklist is active and clears when the last step is done or the checklist is aborted. The `checklist` and `progress_tracker` share data shape so the two never drift.
- **`project` skill** — same treatment for its task list. Could optionally promote to canvas for longer multi-day work where the user wants the panel always available.
- **Future occupants** — `delegate_task` progress, scheduled-task in-flight status, multi-turn approvals.

This means the agent doesn't have to remember to render anything; the existing tools become widget producers as a side effect of doing their job.

## Schema

- `modes`: `["inline", "canvas", "sticky"]` — sticky requires #419
- `accepts_input`: `false`
- `data_schema`:
  - `steps` (array, required) — each `{label, status: "pending"|"in_progress"|"done"|"failed", note?}`
  - `title` (string, optional)

## Notes

- v1: snapshot-rendered. Each update replaces the full step list (no patching).
- Default mode hint: `sticky` when promoted by a workflow tool, `inline` when invoked directly.
- Style: keep the visual quiet — this is status, not the main content.

## References

- Widget catalog epic: #256
- Sticky widget slot (primary surface for the checklist integration): #419
- Existing checklist tool: `src/decafclaw/tools/checklist_tools.py`
- Existing project skill: `src/decafclaw/skills/project/`

Issue #555, opened 2026-05-19
Improve a11y for clickable sidebar list rows (schedules, vault, files)


lmorchard:
## Problem

All three sidebar list components (\`vault-sidebar\`, \`files-sidebar\`, \`schedules-sidebar\`) use plain \`<div @click>\` for clickable rows:

- \`src/decafclaw/web/static/components/vault-sidebar.js:261\` — wiki item rows
- \`src/decafclaw/web/static/components/files-sidebar.js:286\` — file rows
- \`src/decafclaw/web/static/components/schedules-sidebar.js:127\` — schedule rows

This means:
- Not focusable via keyboard (no \`tabindex\`)
- Not announced as interactive by screen readers (no \`role\` or semantic element)
- Activation requires mouse / pointer; Enter/Space don't trigger

The Copilot review on PR #554 flagged this for the new schedules sidebar. The original fix (adding \`role=\"button\"\` + \`tabindex=\"0\"\` + keydown handler) was reverted because Pico v2 applies button styling to anything with \`role=\"button\"\`, which broke the row visual.

## Constraint

Pico v2 styles \`[role=\"button\"]\` like a real \`<button>\` (blue background, primary-inverse text, padding). Adding the attribute to a list row creates a visual regression we'd have to fight CSS-wise. The vault and files sidebars accept the same a11y limitation, so this is project-wide.

## Suggested approaches (pick during brainstorm)

1. **Override Pico's \`[role=\"button\"]\` styles on list rows.** Add CSS reset for \`.conv-item[role=\"button\"]\` / \`.wiki-item[role=\"button\"]\` / \`.schedule-row-header[role=\"button\"]\` that strips background/color/padding back to list-item defaults. Tag-qualify to win the cascade.
2. **Use \`<button>\` elements with full Pico-override CSS.** Most semantically correct. Bigger CSS surface — every button rule needs override.
3. **Use \`tabindex=\"0\"\` + \`@keydown\` but skip \`role=\"button\"\`.** Focusable + operable for keyboard users, but screen readers see a generic focusable region. Add \`aria-label\` for SR users to hear what activates. Compromise; cheapest.

## Scope

- Apply consistent pattern to all three sidebars.
- Visual: must match current appearance.
- A11y verification: tab through sidebar, activate row with Enter/Space, VoiceOver/NVDA announce the row's purpose.

## Priority

[...]

PR #642, opened 2026-07-24
feat(widgets): sticky widget slot above chat input (#419)

### LINKED ISSUE #419: Sticky widget slot above chat-input — workflow-pinned mini-canvas
## Summary

Add a third widget mode — `sticky` — that pins a single widget in a slot directly above the chat-input area for the duration of a workflow. Companion surface to `inline` (scrolls away) and `canvas` (separate panel). Inspired by Claude Code's checklist UI, where in-progress multi-step work stays visible above the prompt instead of scrolling out of view.

## Motivation

When the agent is mid-workflow (running a checklist, executing a delegated task, walking a multi-step plan), the user currently has to scroll up to see progress as new chat content pushes it off-screen. Canvas is the wrong surface — it's for long-lived artifacts, not transient workflow status, and on mobile it's hidden behind a disclosure. Inline is the wrong surface — by design it scrolls away.

A sticky slot answers the question "what is the agent doing right now?" at a glance.

### Surface

- New mode value: `"sticky"` alongside `"inline"` / `"canvas"` in widget descriptors.
- Single-slot — only one sticky widget at a time per conversation. Promoting a new one replaces the previous.
- Position: between the chat scroll area and `chat-input`, full width of the chat column.
- Max height with collapse-to-summary affordance (e.g. 30vh with a `[▾]` toggle to collapse to a one-line summary line).

### Lifetime — tied to a workflow primitive

Don't expose free-form "pin this widget until I clear it" semantics — that gets messy. Instead, lifetime is owned by a workflow primitive:

- **Checklist** (#414, primary first integration): when a checklist becomes active, auto-promote a `progress_tracker` widget into the sticky slot; auto-clear on `checklist_step_done` of the last step, or on `checklist_abort`.
- Followups: `delegate_task` progress, multi-turn approval flows, scheduled task in-flight status.

The agent can also explicitly clear via a `widget_unpin_sticky()` tool if needed, but the default path is workflow-driven.

### Sidecar persistence

Per-conversation `{conv_id}.sticky.json` (or fold into existing `canvas.json` as a `sticky_tab` field — TBD during impl). Survives reload like canvas tabs do.

### WS surface

Minimal extension of the existing canvas WS event kinds — likely a new `sticky_set` / `sticky_clear` pair, or a `target: "sticky"` discriminator on the existing `canvas_update` payload. Pick whichever lands cleanly given the Phase 4 multi-tab refactor.

### Mobile

- Max height tighter than desktop (e.g. 25vh).
- Collapse-to-summary should be the default state on mobile, expanded on desktop.
- Tap the summary line to expand to full widget.

## Open questions

- Should sticky widgets accept input, or display-only? Lean display-only for v1 — input widgets stay inline so the agent's "pause and ask" flow is unambiguous.
- Default mode hint per widget type: `progress_tracker` defaults to `sticky` when promoted by a checklist, but defaults to `inline` if invoked directly. Worth specifying in `widget.json`?

## References

- Widget catalog epic: #256
- First occupant: `progress_tracker` widget #414
- Existing checklist tool: `src/decafclaw/tools/checklist_tools.py`
- Related: #364 (collapse `EndTurnConfirm` into a widget — could share input/sticky-slot infra)

lmorchard:
Adds a third widget surface — **`sticky`** — a single-slot, display-only widget pinned directly above the chat input, companion to `inline` (scrolls away) and `canvas` (detached panel). It answers "what is the agent doing right now?" at a glance while a workflow runs.

This PR is the **infrastructure**; the first real occupant (`progress_tracker`) and the checklist-driven auto-emit are #414, built next on top of this.

## What's here

- **`sticky` widget mode** — opt-in per `widget.json`; enabled on `markdown_document` as the test/demo occupant.
- **`sticky.py`** — per-conversation `sticky.json` sidecar (fail-open reads, atomic writes) + `set_sticky`/`clear_sticky` with mode validation + fail-open emit. Mirrors `canvas.py`.
- **`widget_pin_sticky` / `widget_unpin_sticky`** agent tools (`tools/sticky_tools.py`). Single-slot: a new pin replaces the old.
- **`sticky_set` / `sticky_clear`** WS message types (via `message_types.json` codegen) + forwarders in `websocket.py`.
- **`GET /api/sticky/{conv_id}`** — reload recovery; auth/ownership gate identical to the canvas endpoint.
- **Frontend** — `sticky-state.js` (single-slot state + collapse flag) + `<sticky-slot>` component inserted between `<chat-view>` and `<chat-input>`, with a collapse-to-summary affordance (expanded desktop / collapsed mobile ≤639px) and quiet styling.
- Docs (`docs/widgets.md` "Sticky slot" section) + `CLAUDE.md` key-files.

## Design decisions & divergences from the issue

Explored the code first (see `docs/dev-sessions/2026-07-23-1545-sticky-widget-slot/spec.md`). Key findings/divergences:

- **Explicit `set_sticky` primitive + tools**, not a `target:"sticky"` routing value — the `target` field is effectively vestigial (the web client ignores it; canvas is populated only by explicit tools). Mirroring canvas is the proven path.
- **Ship `widget_pin_sticky`** (the issue leaned against a free-form pin) — single-slot + display-only + explicit unpin bounds the "mess," and it makes this PR self-contained and testable instead of dead until #414. The checklist will drive lifetime in #414 by calling the same `set_sticky`.
- Separate `sticky.json` sidecar + dedicated WS types (both were left TBD in the issue).

## Testing

- Backend unit tests mirroring `test_canvas.py`: sidecar round-trip, fail-open, mode-validation reject, emit assertions, single-slot replace, WS forwarder, REST recovery + auth/ownership. `make check` + `make test` green (3131 passed).
- Built task-by-task with a per-task spec+quality review and a final whole-branch review (all clean).
- **Not yet done: browser smoke test** — the rendered slot (appear/collapse/persist) hasn't been verified in a real browser yet. Backend is fully unit-tested; frontend is `check-js`-clean. Live UI verification is the next step before merge.

Closes #419

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -167,12 +167,13 @@ Full doc index: [docs/index.md](docs/index.md). Hot files for navigation:
 - `compaction_decisions.py` — Structured decision slice threaded through compaction (decisions / open questions / artifacts; persisted to per-conv sidecar)
 - `context_cleanup.py` — Lightweight clear tier: stubs old large tool messages before compaction
 - `canvas.py` — per-conversation canvas state sidecar + state operations
+- `sticky.py` — per-conversation sticky-slot sidecar (`sticky.json`) + `set_sticky`/`clear_sticky`
 - `persistence.py`, `attachments.py`, `embeddings.py`, `frontmatter.py`, `memory_context.py`, `checklist.py`, `notes.py`
 
 ### Tools
 - `tools/tool_registry.py` — Priority-based classification, deferred catalog
 - `tools/search_tools.py` — `tool_search`
-- `tools/{core,workspace_tools,conversation_tools,checklist_tools,notes_tools,shell_tools,http_tools,skill_tools,delegate,model_tools,confirmation,health,attachment_tools,email_tools,heartbeat_tools,canvas_tools}.py`
+- `tools/{core,workspace_tools,conversation_tools,checklist_tools,notes_tools,shell_tools,http_tools,skill_tools,delegate,model_tools,confirmation,health,attachment_tools,email_tools,heartbeat_tools,canvas_tools,sticky_tools}.py`
 - `preempt_search.py` — Keyword-match for pre-emptive tool promotion
 
 ### Skills (bundled)
@@ -192,6 +193,7 @@ Full doc index: [docs/index.md](docs/index.md). Hot files for navigation:
 - `mcp_client.py`
 - `media.py`, `widgets.py`, `widget_input.py`, `web/static/widgets/`, `web/static/components/widgets/widget-host.js`
 - `web/static/components/canvas-panel.js`, `web/static/lib/canvas-state.js`
+- `web/static/components/sticky-slot.js`, `web/static/lib/sticky-state.js` — sticky slot (pinned display-only widget above chat input)
 - `terminals.py` — human-only PTY session registry backing the web terminal ([docs/web-terminal.md](docs/web-terminal.md))
 - `util.py`, `eval/`
 
--- a/docs/dev-sessions/2026-07-23-1545-sticky-widget-slot/plan.md
+++ b/docs/dev-sessions/2026-07-23-1545-sticky-widget-slot/plan.md
@@ -1,0 +1,1086 @@
+# Sticky Widget Slot Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add a single-slot `sticky` widget surface pinned above the chat input, driven by an explicit backend primitive + agent tools, mirroring the canvas subsystem.
+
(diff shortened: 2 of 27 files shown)


lmorchard:
Addressed both in 324166a:

- **Display-only invariant (`sticky.py`):** `set_sticky` now rejects a widget that declares `sticky` mode but also `accepts_input` — good catch, the spec commits to display-only for v1 and validation wasn't enforcing it. Added `test_set_sticky_rejects_input_widget`.
- **`type="button"` on the toggle:** added.

On the **keyboard accessibility of the header `<div>`**: I left it without `role`/`tabindex` deliberately. The header already contains a keyboard-accessible `<button>` (with `aria-label="Toggle sticky panel"`) that performs the collapse/expand, so the control *is* reachable via keyboard/AT — tab to the button, Enter/Space. The `@click` on the surrounding `<div>` is just a larger mouse hit-target, which matches this project's established clickable-row convention (vault/files/schedules sidebars do the same; the cross-cutting a11y fix for those clickable divs is tracked in #555). Adding `role="button"`/`tabindex` to the div would create a second, redundant focus stop for the same action. Happy to revisit if you'd rather fold this into #555's scope.

`make check` + the sticky tests pass.

lmorchard:
Good catch — confirmed and fixed in cee4602. My display-only guard (`324166a`) reads `descriptor.accepts_input`, but `test_http_sticky.py`'s `SimpleNamespace` stub omitted it, so `test_get_sticky_state_pinned` raised `AttributeError`. My mistake was only re-running `test_sticky.py` after that fix instead of the full suite.

The real `WidgetDescriptor` always carries `accepts_input` (dataclass field, default `False`), so the production guard is safe — the stub just needed to mirror it. Added `accepts_input=False` to the stub. Full suite is green again (3132 passed). (The other canvas stubs don't declare `accepts_input` but never hit this guard — canvas validation doesn't read it — verified via the full run.)

The suppressed low-confidence a11y note on the clickable `<div>` is the same one addressed in my previous reply (keyboard path exists via the labelled toggle button; div-click is a mouse convenience per the project's clickable-row convention, cross-cutting fix tracked in #555).

PR #648, opened 2026-07-24
fix(web-terminal): debounce resizes to stop the prompt-reprint storm

lmorchard:
## Problem

Switching conversations spewed a cascade of duplicated shell prompt lines — at *ramping* widths — into the terminal scrollback. (Follow-on from #627, discovered while smoke-testing #647.)

## Root cause (confirmed by instrumented live repro, not inference)

The canvas panel animates its width (`transition: width 0.25s` in `canvas.css`) on every conversation switch. The widget's `ResizeObserver` therefore fires ~once per animation frame, and `_onResize` did an xterm `fit()` + sent a `resize` frame on **every** callback with no coalescing — walking the PTY through every intermediate width. Each genuine `TIOCSWINSZ` makes the shell reprint its prompt, and those reprints accumulate in the ring buffer and re-render on every reconnect+replay.

Evidence gathered by temporarily instrumenting the server:
- A macOS kernel test confirmed identical `TIOCSWINSZ` calls **don't** re-signal — so the storm was genuine size *changes*, not repeats.
- Server logs showed each (re)connect firing `resize cols=2` (a garbage mid-animation width) then `cols=63` — the `cols=2` came from an **immediate** `_onResize()` on `buffer_replay_done` firing mid-animation.

## Fix (client-side, `widget.js`)

- **Debounce** ResizeObserver-driven fits (`_scheduleResize`, 150 ms) → fit + send exactly once, after the layout settles.
- **Dedup** — only send `resize` when the fitted `cols`/`rows` actually changed (hidden panel fits to 0; sub-pixel jitter re-fires the observer).
- Route the **initial post-replay** size through the same debounce instead of an immediate fit (that's what was shipping `cols=2`). Reset the dedup baseline on each `_connect` so a fresh socket still registers its settled size server-side.
- Docs: `docs/web-terminal.md` "Multi-attach and resize" section documents the debounce + why.

## Verification

Instrumented the server and reproduced before/after. **Post-fix:** each (re)connect emits a single `resize` at the settled width, which matches the PTY's existing size → kernel dedups → no reprint; ring buffer holds steady (`2377 → 2377`) across round trips. `make check` / `check-js` clean; terminal unit tests pass.

**Note:** switching *conversations* still tears down + reconnects the socket (by design — the canvas panel unmounts per conversation; keep-alive only spans canvas *tab* switches). Benign now that resizes are clean. No JS unit harness exists for the widget — verification is `check-js` + the instrumented live repro.

### DIFF
--- a/docs/web-terminal.md
+++ b/docs/web-terminal.md
@@ -165,6 +165,19 @@ broadcasts `size_changed` so every client can letterbox if its own viewport
 is larger. This means a small popped-out window can force the in-canvas
 view (and any program relying on terminal size) down to its dimensions.
 
+The widget **debounces** resizes (`_scheduleResize`, 150 ms) and only sends a
+`resize` frame when the fitted `(cols, rows)` actually changed. This matters
+because the canvas panel animates its width (`transition: width 0.25s` in
+`canvas.css`) whenever a conversation is switched, so the widget's
+`ResizeObserver` fires roughly once per animation frame. Without coalescing,
+the terminal would walk the PTY through every intermediate width — and each
+genuine `TIOCSWINSZ` makes the shell reprint its prompt, so a single switch
+produced a cascade of prompt lines at ramping widths in the scrollback. The
+initial post-replay size (sent after `buffer_replay_done`) goes through the
+same debounce, not an immediate fit: a (re)connect lands mid-animation, and an
+immediate fit would grab a garbage intermediate width (e.g. 2 columns) and
+ship it to the PTY.
+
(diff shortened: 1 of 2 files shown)


lmorchard:
Good catch — addressed in ccfaf97. `_send` now returns whether it actually transmitted, and `_onResize` advances the dedup baseline (`_lastCols/_lastRows`) only on a real send. So a debounced resize firing mid-replay no longer poisons the baseline, and the post-`buffer_replay_done` resize is guaranteed to reach the server.

Issue #683, opened 2026-07-25
tool_choice loadout doesn't reproduce production tool deferral (97 callable vs 40 active + 58 prose-listed)


lmorchard:
Found while diagnosing #670. **This was not the cause of #670** — both tools in that investigation were active in both harnesses — but it is a real fidelity gap in how we measure tool disambiguation, and it should be a deliberate choice rather than an accident.

## The gap

`tool_choice` presents every tool as callable. The real agent's first turn does not.

Measured in a worktree off `00334d7` (`tmp/probe_loadout.py`, no LLM calls — just `build_tool_list(ctx)` vs `build_full_tool_loadout(config)` under an eval-shaped config):

| | `tool_choice` | agent, turn 1 |
|---|---|---|
| callable tool schemas | **97** | **40** |
| `<deferred_tools>` prose listing | none | **3,173 chars, 58 tools** |
| `tool_search` offered | no | yes |

So 58 tools are directly callable in one harness and only prose-summarized-behind-`tool_search` in the other. Notably `tool_search` itself is the one tool present for the agent but absent from the `tool_choice` loadout.

The docstring in `src/decafclaw/eval/tool_choice/loadout.py:1-12` is explicit that this is intentional:

> Builds a fully-loaded tool definitions list … without the production deferral or activation gating logic. The eval intentionally measures description overlap under fair conditions, so every tool the model could pick is in scope simultaneously.

That is a defensible design for measuring *description overlap*. It's a poor proxy for *what the agent will actually do*, because in production a `normal`-priority tool competing with a `critical` one isn't in the schema at all — it's a one-line prose summary the model must first `tool_search` for.

## Why it matters

Any `tool_choice` case whose `expected` or `near_miss` tool is deferred in production is measuring a decision the agent never gets to make in that form. The pass tells you the descriptions disambiguate; it doesn't tell you the routing works.

Concretely, of the 58 deferred-at-turn-1 tools, several appear in `evals/tool_choice/core_overlaps.yaml` cases — e.g. `conversation_search`, `conversation_compact`, `workspace_write`/`workspace_append`, `canvas_*`, `send_email`, `http_request`.

## Possible directions

Not proposing one — that's the work.

1. **Leave as-is, document the limitation.** Cheapest. `tool_choice` stays a pure description-overlap instrument and we stop reading it as a routing predictor.
[...]

Issue #676, opened 2026-07-25
make eval-tools has 4 failing disambiguation cases


lmorchard:
Found while investigating #670. `make eval-tools` (~30s) currently fails four cases. Filing separately since they're unrelated to each other and to #670.

```
FAIL  workspace-write-vs-canvas-save-blog-post      picked <no_tool>;   expected workspace_write
FAIL  tabstack-automate-vs-research-form-fill       picked web_fetch;   expected tabstack_automate
FAIL  tabstack-research-vs-automate-multi-source    picked web_fetch;   expected tabstack_research
FAIL  vault-recent-vs-list-recent-journal           picked <no_tool>;   expected vault_recent
```

Two distinct shapes:

**Picked `<no_tool>`** (`workspace-write-vs-canvas-save-blog-post`, `vault-recent-vs-list-recent-journal`) — the model declined to act at all rather than choosing wrong. That's usually a sign the scenario reads as conversational, or the tool's description doesn't claim the case. Different failure from a mis-route and probably a different fix.

**`tabstack_*` losing to `web_fetch`** — both tabstack cases went to `web_fetch`. `web_fetch` is likely winning on generality; the tabstack descriptions may not assert their differentiators (automation/multi-source research) strongly enough. These two probably share one fix.

## Context

`make eval-tools` is cheap and fast — ~30s, no meaningful token cost compared to the full suite. The reason these went unnoticed is the same reason everything else surfaced on 2026-07-24: nothing had been run. The last recorded full eval run before today was 2026-05-31.

Worth considering whether `make eval-tools` should be wired into a pre-merge check given its cost profile, which would make this class of drift a CI failure instead of an archaeology finding. That's a separate discussion from fixing these four.

Size: S.

Issue #671, opened 2026-07-24
vault_section: section paths must be rooted at the page H1, so bare heading names never resolve


lmorchard:
Found by the first full `make eval` run in ~8 weeks (2026-07-24, 45/52). Eval case *"adds a section without rewriting other sections"* failed on `Too many tool errors: 2 > 1`, both errors being:

```
[error: target section not found]
```

I reproduced it deterministically — this is a hard failure, not the agent fumbling:

```python
text = "# Project Notes\n\n## Background\n\n...\n\n## TODO\n\n- Old item\n..."

add_section("Status", level=2, after="Background")     -> False
add_section("Status", level=2, after="## Background")   -> False
add_section("Status", level=1, after="Background")      -> False
add_section("Status", level=2, after="background")      -> False
```

Every spelling fails.

## Root cause

`find_section` resolves **slash-delimited paths rooted at the page's H1**, and the H1 counts as a path segment:

```python
find_section("Background")                  -> None
find_section("Project Notes/Background")    -> Section(title='Background', level=2, ...)
find_section("TODO")                        -> None
```

So on any page with an H1 title — i.e. essentially every vault page — a `##` section is only addressable as `<H1 Title>/<Section>`.

## Why the schema doesn't save it

The param descriptions do say "slash-separated":

```
after: Slash-separated section path to insert/move after (e.g. 'top/first'). Used by add and move.
```

But nothing conveys that the *first segment must be the page's H1*. Read plainly, `'top/first'` suggests a nesting path among sections, and a bare top-level `## Background` reasonably looks like the one-segment path `Background`. The one thing a caller needs to know is the one thing that's missing — and the tool returns a flat `target section not found`, which reads as "no such heading" rather than "your path is under-qualified."

## Fix (preference order)

1. **Resolve a bare title when it's unambiguous.** If `find_section(name)` misses, fall back to matching a unique section whose title equals `name`, and only fail on genuine ambiguity. This is the ergonomic fix and it stops the LLM from having to know a non-obvious convention.
2. **Make the error actionable.** `[error: target section not found: 'Background'. Known paths: Project Notes, Project Notes/Background, Project Notes/TODO, ...]` turns a dead end into a self-correcting retry.
3. **Fix the descriptions regardless** — state that paths are rooted at the page title, with a realistic example (`'Project Notes/Background'`, not `'top/first'`).

(1) and (2) compose well: the fallback handles the common case, the improved error handles the ambiguous one.

## Note on the eval

[...]

Issue #650, opened 2026-07-24
eval harness: allow arbitrary config overrides so harness configurations can be A/B'd


lmorchard:
## Problem

The eval runner can override exactly three things:

- `setup.max_tool_iterations` → `config.agent.max_tool_iterations` (`eval/runner.py:490`)
- `setup.reflection_enabled` → `config.reflection.enabled` (`eval/runner.py:491`)
- `--model` → `config.llm.model` (`eval/runner.py:711`)

Everything else is pinned to whatever `data/{agent_id}/config.json` resolves to at run time.

That means the large surface of *harness* knobs in `config_types.py` has never been measured against itself. All of these were chosen by judgment and have no empirical backing:

- `vault_retrieval.mode` — `always` vs `headlines` vs `on_demand`, plus the `w_similarity` / `w_recency` / `w_importance` weights
- tool deferral thresholds and `preempt_search` keyword matching
- `cleanup` tool-result clearing thresholds and the protected-window size
- `compaction.decisions_enabled`, `decisions_max_per_category`, memory-sweep on/off
- `notes.context_max_entries` / `context_max_chars`
- `recent_journal.max_hours` / `max_entries` / `max_tokens`

## Why now

Reading <https://ahmadrosid.com/blog/what-if-the-harness-mattered-more-than-the-model>. Its central evidence is HarnessBench: 106 tasks, two harness configurations over the same model, 52.4% vs 76.2%. A 24-point spread from configuration alone — larger than most model upgrades buy.

We have unusually good measurement infrastructure for a project this size (three telemetry streams, an eval harness with trend history). But every one of those measures the *model and prompts* under a fixed harness. The one variable the post says dominates is the one we can't vary.

---

## Revised plan (statistical power)

The original write-up proposed config overrides plus a sweep mode as one unit. Sizing the suite first changes that: **the eval suite is 51 cases total.**

| file | cases | | file | cases |
|---|--:|---|---|--:|
| memory.yaml | 8 | | tool-deferral.yaml | 4 |
| project-skill.yaml | 8 | | conversation.yaml | 3 |
| vault.yaml | 6 | | delegate.yaml | 3 |
| memory-multi-turn.yaml | 5 | | shell.yaml | 2 |
| workspace-tools.yaml | 4 | | 9 files × 1 | 9 |

[...]

PR #689, opened 2026-07-25
Eval runner: assemble the system prompt (45/52 → 53/54)

### LINKED ISSUE #670: notes_append vs vault_journal_append: tool_choice and the full agent loop disagree on the same prompt
Found by the first full `make eval` run in ~8 weeks (2026-07-24, 45/52). **Two independent cases failed identically**, so this is not a one-off sample:

```
[20/52] saves memory when asked
  Expected one of ['vault_journal_append'] to be called, but tools called were [notes_append]

[43/52] saves user-level fact to vault journal on 'remember'
  Expected one of ['vault_journal_append'] to be called, but tools called were [notes_append]
```

## Why this matters beyond a red test

`notes_append` writes to the **per-conversation** scratchpad (`conversations/{id}/notes.md`). `vault_journal_append` writes **durable, cross-conversation** episodic memory.

So when the user says "remember that X", the fact is being recorded somewhere it will not survive into any other conversation. The agent will confirm it saved, and the memory is effectively lost at conversation boundaries. That's a silent data-loss-shaped bug, not cosmetic.

## Root cause: the descriptions collide, and the wrong one is louder

```
notes_append:
  "Jot a single note to the per-conversation scratchpad — use for things you want
   to remember **across turns within this conversation**: 'user said X', 'we
   decided Y', ... **Prefer this over `vault_write` for transient
   conversation-scoped facts** ..."

vault_journal_append:
  "Save a journal entry — a timestamped observation, fact, or note. ... Use this
   for anything worth recording: user preferences, facts, project details,
   decisions, conversation context."
```

Three things stack against the right answer:

1. **`notes_append` literally names the trigger phrase.** `'user said X'` is what "remember that my favorite color is blue" *is*.
2. **`notes_append` disambiguates against the wrong tool.** It says "prefer this over `vault_write`" — never over `vault_journal_append`, which is the actual collision.
3. **`notes_append` is emphatic, `vault_journal_append` is flat.** Bold plus "Prefer this over" vs. a plain list. Per CLAUDE.md, *"Tool descriptions are a control surface. Wording changes … measurably change LLM behavior."* The louder description is winning, and it happens to be the wrong one.

`vault_journal_append` also claims overlapping ground ("conversation context"), which doesn't help.

## This was predicted

CLAUDE.md, verbatim:

> PR #429's smoke test bit-rotted in three weeks because `notes_append` arrived later with no `tool_choice` case guarding the disambiguation — that's the rot vector this convention catches.

The convention was right and the guard was never added. Two eval cases now fail on exactly that seam.

## Fix

1. **Sharpen both descriptions on the durability axis** — the distinction is lifetime, not content type. `notes_append` should disambiguate explicitly against `vault_journal_append` (not just `vault_write`), and should stop using `'user said X'` as its example since that's the journal's job. `vault_journal_append` should assert the "remember/save this" trigger.
2. **Add the missing `tool_choice` case** for `notes_append` vs `vault_journal_append` in `evals/tool_choice/`. `make eval-tools` is ~30s, so this guard is nearly free and is what would have caught the drift 8 weeks ago.
3. Re-run cases 20 and 43 to confirm.

Size: S. Worth doing before #650's noise-floor work, since these two are deterministic failures that would otherwise pollute a variance measurement.

lmorchard:
## Summary

The full-agent eval harness has been running **every case with an empty system prompt**. `config.system_prompt` is assembled in exactly one place — `decafclaw/__init__.py`, the app entry point — and the eval CLI calls `load_config()` + `run_eval()` directly, never going through it. `config.py` only reads the field from a `SYSTEM_PROMPT` env var (default `""`).

`ContextComposer._compose_system_prompt` reads that field verbatim, so eval turns got a **zero-length system message**: no SOUL.md, no AGENT.md, no `<skill_catalog>`, and no always-loaded skill bodies — including the vault SKILL.md that documents when to journal.

`tool_choice` doesn't share the gap (it calls `load_system_prompt` directly). That is the entire harness disagreement reported in #670 — and it **inverts** the issue's leading hypothesis: `tool_choice` was the *higher*-fidelity harness; the full-agent runner was the broken one.

## Result: 45/52 → 53/54

| | baseline `1621` | with extra skills `1822` | hermetic `2323` |
|---|---|---|---|
| result | **45 / 52** (86.5%) | 51 / 52 (98.1%) | **53 / 54** (98.1%) |
| duration | 647.2s | 473.0s | 605.8s |
| tokens | 1,308,185 | 2,272,538 | 2,235,913 |

54 cases rather than 52 because the rebase brought in two new ones (#681, #677); both pass. The middle column is the first re-baseline, before review caught the `extra_skill_paths` hole — see "Review round" below.

**All seven baseline failures flipped green:**

| case | baseline | new |
|---|---|---|
| recalls recent memories (project + style) | fail | **pass** |
| saves memory when asked *(#670 case 20)* | fail | **pass** |
| postmortem produces all five sections, blamelessly framed | fail | **pass** |
| writes spec when asked | fail | **pass** |
| dream consolidation fills page frontmatter after writing | fail | **pass** |
| saves user-level fact to vault journal on 'remember' *(#670 case 43)* | fail | **pass** |
| adds a section without rewriting other sections | fail | **pass** |

**Six of the seven "failures" in the 45/52 baseline were harness artifacts, not agent defects.** Anyone who had picked up one of those issues would have been debugging an agent that doesn't ship.

## Evidence

Measured on `vertex-gemini-flash`, `expect_tool: vault_journal_append`:

| condition | case-20 prompt | case-43 prompt |
|---|---|---|
| as shipped (empty prompt) | **1/15** | **13/15** |
| system prompt assembled | **10/10** | **10/10** |
| `tool_choice` | **5/5** | **5/5** |

Both of the issue's cheap-to-rule-out hypotheses were ruled out by measurement, not reasoning:

- **Trailing "Confirm what you saved."** — a 2×2 crossing moved 1/5 → 0/5 and 5/5 → 4/5. Both within noise; *content* drives the per-case split, not the clause.
- **Reflection retries** — `reflection.enabled: false`, 3 reps: 0/3, still `notes_append`, 1 tool call each.

Also corrects the issue's framing: **case 43 is flaky (13/15), not deterministic.** "Two cases failed identically, so not a one-off" holds for case 20 (0/15) only.

## Design decisions

- **Assembly happens in `run_test`, after `_build_test_config` applies the sandbox.** The sandbox clears *two* sources of machine-local state so the prompt resolves to the bundled tier only:

  | sandbox field | excludes |
  |---|---|
  | `agent.data_home` → tmp | `data/{agent_id}` prompt overrides, incl. USER.md |
  | `extra_skill_paths` → `[]` | skills from external dirs (`~/.agents/skills`) |

  Both applied last, so `setup.config_overrides` can't opt back into either. Rejected: assembling in `__main__.py` against the un-sandboxed config (higher fidelity to one deployed agent, but machine-dependent numbers).
- **Guarded on falsiness**, so an explicit `SYSTEM_PROMPT` env override still wins. Covered by a test.
- **`skill_tool_owners` moved under its own guard.** `load_system_prompt` populates `discovered_skills` as a side effect, so folding it into the `if not config.discovered_skills` branch would skip it and silently break `/foo` command dispatch in evals. Hit this during the probe.
- **No tool-description changes.** With the prompt present, routing is 20/20 correct — there is no description defect. The issue body proposed rewording `notes_append` / `vault_journal_append`; that is measurably unnecessary and would have risked the passing control cases.

## Testing

- `tests/test_eval_system_prompt.py` — **7 tests**, no LLM calls (`run_agent_turn` patched at the seam): prompt is non-empty at turn time; it carries `<skill_catalog>` / `<loaded_skills>` / `<skill name="vault">`; `skill_tool_owners` survives; an explicit prompt isn't clobbered; the sandbox clears `extra_skill_paths`; `config_overrides` can't restore it; every discovered skill is `trust_tier == "bundled"`.
- TDD: the first two failed before the fix with `AssertionError: assert ''`. The env-override test was verified to have teeth by making the guard unconditional (fails), then restoring.
- `make test` **3449 passed, 2 skipped**. `make check` clean (ruff, pyright 0 errors, tsc, message-types drift).
- The three `vault_journal_append` ↔ `notes_append` control cases in `evals/tool_choice/core_overlaps.yaml` **pass, untouched** — pair overlap 0/1 and 0/2 swapped (0%).

## Things reviewers should know

**Two noise-floor failures surfaced, neither a regression.** Both were checked by re-running in isolation rather than assumed:

- `continues interviewing after first answer` failed in the `1822` run (answering *"Project ... has been created. Please call `project_next_task()`"* — skipping the interview and instructing the **user** to call a tool). Re-ran turn 1: **3/3 PASS**. It also passes in the final `2323` run.
- `uses workspace_move for a rename` is the single failure in the final run. Re-ran: **3/4 PASS**. The agent reaches for a vault tool on a path-shaped rename and gives up on `[error: vault page 'drafts/old-name.md' not found]` — a `vault_*` vs `workspace_*` disambiguation wobble.

Both recorded in `notes.md`, neither filed; they belong to #650 / #683.

**#671 may be a harness artifact.** Its eval case (`adds a section without rewriting other sections`) now passes. Worth re-checking before anyone starts on it. Same for the `dream` frontmatter case. Not investigated here.

**#676's "4 failures" is a snapshot, not a set.** Across three `make eval-tools` runs the failure set was 4 / 6 / 4, with only the two tabstack cases (`tabstack-automate-vs-research-form-fill`, `tabstack-research-vs-automate-multi-source`) failing every time. This PR can't affect `tool_choice` — it has its own runner that never imports `eval/runner.py` — so this is pre-existing instability. #676 likely wants re-scoping to those two, with the rest folded into #650.

**Two adjacent changes, both flagged:**
- Corrected a stale claim in `docs/eval-loop.md` that `evals/history.jsonl` is "committed to git" — it has been gitignored since #606. The new history-discontinuity note references that file, so leaving the contradiction beside it would have been worse.
- Added `tmp/` to `.gitignore` (untracked, wasn't ignored).

## Review round

Copilot caught a real hole, flagged on both the code comment and the docs: `load_system_prompt` calls `discover_skills(config)`, which reads `config.extra_skill_paths` — and those live *outside* `data_home`, so the tmp sandbox never reached them. The "bundled tier only" claim was false.

Verified and quantified rather than taken on faith:

```
extra_skill_paths = ['~/.agents/skills']
tiers: Counter({'extra': 113, 'bundled': 12})

WITH extra_skill_paths: 34,704 chars, 125 skills
WITHOUT:                23,262 chars,  12 skills
-> 11,442 chars, 33% of the eval system prompt
```

A third of every eval prompt was one developer's personal skill catalog. No skill *bodies* leaked (none are always-loaded), but all 113 catalog entries did — so eval results were machine-dependent, exactly what the design decision claimed to prevent.

**Fixed** by clearing `extra_skill_paths` in the sandbox, plus three tests and corrected comment/docs wording. The suite was then re-baselined a second time under the hermetic config, which is the `2323` column above.

## References

- Closes #670
- Session docs: [`docs/dev-sessions/2026-07-24-1728-670-notes-vs-vault-harness-fidelity/`](docs/dev-sessions/2026-07-24-1728-670-notes-vs-vault-harness-fidelity/) — `spec.md`, `research.md` (full measurement tables + hypothesis verdicts), `plan.md`, `notes.md` (flip table)
- Follow-up filed during diagnosis: #683 — `tool_choice` loadout doesn't reproduce production tool deferral (97 callable vs 40 active + 58 prose-listed). Did **not** cause #670; independent of this fix.
- Related, not addressed: #671, #676, #650

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,6 @@ data/
 evals/results/
 evals/history.jsonl
 node_modules/
+tmp/
 .playwright-mcp/
 .claude/
--- a/docs/dev-sessions/2026-07-24-1728-670-notes-vs-vault-harness-fidelity/notes.md
+++ b/docs/dev-sessions/2026-07-24-1728-670-notes-vs-vault-harness-fidelity/notes.md
@@ -1,0 +1,211 @@
+# Notes — #670 eval runner system-prompt fidelity
+
+## Phase 3 — targeted eval re-run (2026-07-24)
+
+Model `vertex-gemini-flash`, run in this worktree at commit `0a4073f`.
+
+### `evals/memory.yaml` — 7/8
+
+```
+[1/8] finds preference by direct term                        PASS  (2.0s, 13962 tokens, 0 tools)
+[2/8] finds preference by related term                       PASS  (2.3s, 13961 tokens, 0 tools)
+[3/8] recalls recent memories (project + style)              FAIL  (16.5s, 28005 tokens, 0 tools)
+[4/8] handles missing memories gracefully                    PASS  (2.5s, 13830 tokens, 0 tools)
+[5/8] saves memory when asked                                PASS  (4.3s, 28590 tokens, 1 tools)
+[6/8] connects related memories (event bus + async)          PASS  (2.7s, 14149 tokens, 0 tools)
+[7/8] synthesizes from multiple memories for a complex q.    PASS  (4.9s, 14115 tokens, 0 tools)
+[8/8] finds info with indirect phrasing                      PASS  (2.3s, 13972 tokens, 0 tools)
+8 tests, 7 passed, 1 failed (37.5s, 140584 tokens)
+```
+
+**Case 20 "saves memory when asked" → PASS.** This is #670's primary case.
+
+`recalls recent memories (project + style)` also fails, but it was **already
+failing in the 2026-07-24 baseline** (`evals/results/2026-07-24-1621-vertex-gemini-flash/`)
+— not a flip caused by this change. Left alone per spec.
+
+### `evals/vault.yaml` — 6/6
+
+```
+[1/6] saves user-level fact to vault journal on 'remember'   PASS  (4.3s, 28654 tokens, 1 tools)
+[2/6] reaches for vault_search on explicit search request    PASS  (7.8s, 30578 tokens, 1 tools)
+[3/6] finds specific fact under heavy distractor load        PASS  (2.3s, 13952 tokens, 0 tools)
+[4/6] reads a named vault page directly without searching    PASS  (4.1s, 30216 tokens, 1 tools)
+[5/6] reaches for vault_backlinks to find inbound links      PASS  (3.8s, 29557 tokens, 1 tools)
+[6/6] adds a section without rewriting other sections        PASS  (12.9s, 45059 tokens, 2 tools)
+6 tests, 6 passed, 0 failed (35.2s, 178016 tokens)
+```
+
+**Case 43 → PASS.** Bonus: `adds a section without rewriting other sections`
+was a baseline failure and is now green too — that is the #671-adjacent case,
+so #671 may be partly a symptom of the same missing prompt. **Not investigated
+here; worth a look when #671 is picked up.**
+
+### `make eval-tools` — the three control cases hold
+
+All three `vault_journal_append` <-> `notes_append` cases PASS in both runs,
+and the pair overlap stayed clean:
+
+```
+notes_append <-> vault_journal_append     0/1 swapped (0%)
+vault_journal_append <-> notes_append     0/2 swapped (0%)
+```
+
+### Plan checkbox that did NOT hold as written
+
+> "the four #676 failures are the *only* other failures — no new ones"
+
+**False as written, but not a regression from this change.** `tool_choice` has
+its own runner (`eval/tool_choice/runner.py`) that never imports
+`eval/runner.py`, so nothing in this PR can affect it. What the runs show is
+that **`eval-tools`' failure set is not stable run-to-run**:
+
+| run | failures | count |
+|---|---|---|
+| Les's, in #670 (2026-07-24) | workspace-write-vs-canvas, tabstack-automate, tabstack-research, vault-recent-vs-list | 4 |
+| mine, run 1 | workspace-write-vs-canvas, tabstack-automate, tabstack-research, ask-choice-vs-text, frontmatter-vs-write, no-tool-greeting | 6 |
+| mine, run 2 | canvas-vs-workspace-write, tabstack-automate, tabstack-research, ask-choice-vs-text | 4 |
+
+Summaries: 26/32 (81%) then 28/32 (88%).
+
+Only **two** cases fail in all three runs: `tabstack-automate-vs-research-form-fill`
+and `tabstack-research-vs-automate-multi-source` (both pick `web_fetch`). The
+rest move around, and several `<no_tool>` picks look like the model declining
+to act rather than misrouting.
+
+**Implication for #676:** its "4 failures" is a snapshot, not a set. #676 should
+be re-scoped to the two deterministic tabstack cases, with the rest folded into
+#650's noise-floor work. Not doing that here — out of scope — but it's the
+finding.
+
+---
+
+## Phase 4 — full-suite re-baseline (2026-07-24)
+
+| | baseline `2026-07-24-1621` | new `2026-07-24-1822` |
+|---|---|---|
+| result | **45 / 52** (86.5%) | **51 / 52** (98.1%) |
+| duration | 647.2s | 473.0s |
+| tokens | 1,308,185 | 2,272,538 (**1.74x**) |
+
+Token growth came in at 1.74x, below the 2.4x/case I extrapolated from the
+two-case probe — many cases are multi-turn or tool-heavy, so the fixed ~16.6k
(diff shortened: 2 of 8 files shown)


lmorchard:
Both Copilot comments were the same real issue, and it was worth catching — thanks.

**Confirmed and fixed.** `load_system_prompt` → `discover_skills(config)` → `config.extra_skill_paths`, and those paths live outside `data_home`, so the tmp sandbox never reached them. The "bundled tier only" claim in both the comment and the docs was wrong.

Quantified before fixing:

```
extra_skill_paths = [~/.agents/skills]
tiers: Counter({extra: 113, bundled: 12})

WITH extra_skill_paths: 34,704 chars, 125 skills
WITHOUT:                23,262 chars,  12 skills
-> 11,442 chars, 33% of the eval system prompt
```

So a third of every eval prompt was one developer's personal skill catalog. No skill *bodies* leaked (none of the 113 are always-loaded), but every catalog entry did — which made eval results machine-dependent, precisely the property the change was supposed to deliver.

Rather than only rewording, `_build_test_config` now clears `extra_skill_paths` alongside the `data_home` / `agent.id` sandbox, applied last so `setup.config_overrides` cannot opt back in. Three tests added: the sandbox clears it, `config_overrides` cannot restore it, and end-to-end every discovered skill is `trust_tier == "bundled"`.

The suite was re-baselined a second time under the hermetic config — **53/54** — and the comment and docs now describe both halves of the sandbox in a table instead of implying `data_home` covers it.

PR #705, opened 2026-07-26
vault_section: resolve bare and partial section paths

### LINKED ISSUE #671: vault_section: section paths must be rooted at the page H1, so bare heading names never resolve

lmorchard:
## Summary

`find_section` walked strictly from the document's top-level sections. An H1 page has exactly one of those — the H1 — so every `##` heading was addressable **only** as `<H1 Title>/<Section>`. Nothing in the schema conveyed that, and a caller had no way to guess it:

```
find_section('Background')                   -> None
find_section('Project Notes/Background')     -> Background
add_section('Status', 2, after='Background') -> False
```

The error was a flat `target section not found` — which reads as "no such heading" rather than "your path is under-qualified", and didn't even echo the path.

## Resolution

Exact rooted path first, then a unique **suffix** match against every section's full path. `Background`, `Archive/Background`, and `Project Notes/Archive/Background` all reach the same section, so qualifying a path *further* never makes it stop working. `normalize_title` also strips leading heading hashes, so `## Background` matches.

Ambiguous paths resolve to `None` rather than picking a winner — these back mutating operations, and silently editing the wrong `## Notes` is worse than a failed call.

**All 13 call sites route through `find_section`**, so `add` / `remove` / `rename` / `move` / checklist ops were all fixed by that one change. None of them needed touching.

## Diagnostics

`section_candidates()` and `all_section_paths()` let the error sites tell ambiguous from missing, via a shared `describe_section_miss`. It returns an unwrapped message so `tools.py` and `_insert_into_doc`'s bare-string return share one wording. Paths render with real titles, and a test asserts every candidate offered resolves when pasted back.

```
[error: ambiguous section path 'Background' matches 2 sections:
  Project Notes/Background
  Project Notes/Archive/Background
Use a longer path to disambiguate.]

[error: section not found: 'Statuz'. Known paths:
  Project Notes
  Project Notes/Background
  ...]
```

Duplicate headings get a distinct message: when two sections share a path, "use a longer path" is advice that **cannot be followed** — observed sending the agent into a retry loop until it blew its tool budget.

## Two further defects, found by running the eval rather than the tests

Fixing the paths let the agent reach the next problem, twice over. Both are pre-existing on `main`; the path fix just made them reachable.

**1. `vault_section action=add` could create a section but not fill it** — though `Document.add_section` has always taken `content`. So "add a section titled X with body Y" needed a second operation, and that operation is where it went wrong: one run guessed `content=` and got a `TypeError`, another rewrote the whole page and mangled the layout.

**2. `level` defaulted to 1**, so adding a section after a `##` inserted a second H1 mid-page — which silently reparents everything below it:

```
# Project Notes            # Project Notes
  ## Background              ## Background
  ## TODO             ->   # Status
  ## References              ## TODO         <- reparented
                             ## References   <- reparented

'Project Notes/TODO' silently becomes 'Status/TODO'
```

`level` now defaults to what the anchor implies — sibling of `after`/`before`, one below `parent`, 1 only when unanchored. Explicit levels still win.

Worth being explicit: **exposing `content` alone made the eval repro *worse*** (2/4 → 0/3), by making a one-call path viable and so exposing that default underneath it. Fixing the default is what turned it around.

## Measured

Eval repro across the session, `vertex-gemini-flash`:

| stage | repro | notes |
|---|---|---|
| baseline `2026-07-24-2323` | pass | 4 tool calls |
| after path resolution | 2/4 | bare titles accepted; blocked on `content` |
| after `content` | 0/3 | hit the `level=1` default |
| after level inference | **5/6** | best runs at **1–2 calls** |

Not yet a *reliable* case — it is a flaky corner of the suite and was before any of this — but the defect is fixed and the good path is half the calls.

## Testing

- `tests/test_vault_sections_helpers.py` — 22 new: suffix resolution, exact-wins, ambiguity, `#` tolerance, empty paths, candidate/known-path diagnostics, truncation, duplicate-heading message, level inference, tree integrity.
- `tests/test_vault_section_tools.py` — 6 new at the tool layer.
- TDD throughout; each phase's tests were run red first.
- `make test` **3555 passed, 2 skipped**. `make check` clean.

**One bug I introduced and the unit tests missed:** the tool guarded `not isinstance(level, int)` before `level` became optional, so omitting it died with `level must be between 1 and 6, got None`. The `Document`-level tests call `add_section` directly and sailed straight past the tool-layer guard — only the eval caught it. Two tool-level tests added.

## Things reviewers should know

**No `tool_choice` eval case.** CLAUDE.md asks for one on a sharpened tool description, but that convention targets disambiguation *between* overlapping tools — which tool the model picks. These are parameter descriptions inside a single tool, and a `tool_choice` case asserts on tool name only; it cannot observe an argument value. The eval repro above covers the behavior instead.

**Accepted risk:** `normalize_title` now strips `#` from stored titles too, so a heading literally named `## #hashtag` normalizes to `hashtag`. It applies symmetrically to stored titles and queries, so lookups stay consistent; only a page with two headings differing *solely* by a leading `#` would newly collide.

**Not fixed, noted:** `add_section` emits `## Status\nWorking on it.` with no blank line after the heading. Valid markdown, not idiomatic. Left alone — it touches shared insertion formatting that existing round-trip tests may pin, and it isn't blocking anything.

**Unrelated trap found while preparing this PR:** `make check` mutates `src/decafclaw/web/static/package-lock.json`, deleting 512 lines (the cross-platform optional `esbuild` entries). It got swept into a commit here via `git add -A src/` and had to be backed out. Since #703 just repaired that lockfile, the next person to run `make check` and stage broadly will silently re-break it. Filing separately.

## References

- Closes #671
- Session docs: [`docs/dev-sessions/2026-07-25-0920-671-vault-section-paths/`](docs/dev-sessions/2026-07-25-0920-671-vault-section-paths/) — `research.md` (call-site inventory, deterministic repro), `plan.md`, `notes.md` (per-phase measurements)
- Context: found during #670's full-suite re-baseline. My #670 retro speculated #671 might be a harness artifact because its eval case started passing — that was **wrong**, and this PR's `research.md` records the check that disproved it.

### DIFF
--- a/docs/dev-sessions/2026-07-25-0920-671-vault-section-paths/notes.md
+++ b/docs/dev-sessions/2026-07-25-0920-671-vault-section-paths/notes.md
@@ -1,0 +1,152 @@
+# Notes — #671 vault_section path resolution
+
+## Phase 4 — does the agent actually benefit? (2026-07-25)
+
+`vertex-gemini-flash`, worktree at `6513b5f`.
+
+### `evals/vault.yaml` — 4/6
+
+```
+[1/6] saves user-level fact to vault journal on 'remember'   FAIL  (1 tools)
+[2/6] reaches for vault_search on explicit search request    PASS  (1 tools)
+[3/6] finds specific fact under heavy distractor load        PASS  (0 tools)
+[4/6] reads a named vault page directly without searching    PASS  (1 tools)
+[5/6] reaches for vault_backlinks to find inbound links      PASS  (1 tools)
+[6/6] adds a section without rewriting other sections        FAIL  (3 tools)
+```
+
+Case 1 picked `vault_write` over `vault_journal_append` — the #670 case, which
+scored 10/10 under repeated sampling two days ago. One sample; noise-floor
+territory, and nothing in this branch touches those two descriptions.
+
+### The target case: the fix works, a different defect took over
+
+**Tool calls dropped from 4 to 2 on passing runs** (baseline: 4, in
+`2026-07-24-2323`). That is the win the fix was aimed at.
+
+Re-ran the case 3× per the plan rather than concluding from one run:
+
+| run | result | calls |
+|---|---|---|
+| in `vault.yaml` | FAIL | 3 |
+| `addsec-1` | **PASS** | **2** |
+| `addsec-2` | FAIL | 4 |
+| `addsec-3` | **PASS** | 4 |
+
+2/4 overall. **In every failure `after: "Background"` — the bare title — was
+accepted.** Path resolution is no longer the blocker; #671's defect is fixed.
+
+### What blocks it now: `vault_section add` can't set a body
+
+`Document.add_section` has taken a `content` parameter all along.
+`tool_vault_section` never exposed it — pre-existing on `origin/main`, nothing
+to do with this branch. So "add a section titled X with body Y" always needs a
+second operation, and the second operation is where it goes wrong:
+
+```
+# run 1 — guessed the obvious API
+vault_section {"title": "Status", "after": "Background", "content": "Working on it."}
+  -> [error executing vault_section: tool_vault_section() got an unexpected
+      keyword argument 'content'. Expected parameters: page, action, section,
+      title, level, after, before, parent]
+  -> fell back to read + manual rewrite
+
+# addsec-2 — created the section, then rewrote the page to fill it
+vault_section {"title": "Status", "after": "Background"}   # empty section
+vault_read
+vault_write  # full-page rewrite, mangled the layout
+```
+
+The capability exists one layer down:
+
+```python
+d.add_section('Status', level=2, content='Working on it.', after='Background')
+```
+
+produces exactly the output the eval asserts, in one operation.
+
+**Decision:** Les scoped this in as Phase 5 rather than deferring it — same
+tool, same class of defect (a natural operation the tool blocks), and the
+underlying method already implements it.
+
+### Minor formatting nit — noted, not fixed
+
+`add_section` emits `## Status\nWorking on it.` with no blank line after the
+heading. Valid markdown, not idiomatic. Left alone deliberately: it touches
+shared insertion formatting that existing round-trip tests may pin, and it
+isn't blocking anything.
+
+---
+
+## Phases 5–6 — the content gap, and the level default it exposed
+
+### Phase 5: `content` on `vault_section add`
+
+One parameter, a passthrough, a description, two tests. The eval repro then
+went **2/4 → 0/3** — worse, not better.
+
+That was a genuine (if indirect) regression from the change: by making the
+one-call path viable, it made the agent hit a pre-existing bad default far more
+often.
+
+### Phase 6: `level` defaulted to 1, which reparents the page
+
+The agent omitted `level`, the tool defaulted to `1`, and a `# Status` landed
+mid-page after a `##`. That isn't cosmetic — it silently restructures the
+document:
+
+```
+# Project Notes
+  ## Background
+# Status              <- inserted here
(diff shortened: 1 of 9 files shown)


lmorchard:
Both comments were right — thanks. Fixed in `c7f2a31`.

**1. Mixed anchor precedence — a real regression from my refactor.** Confirmed exactly as described: anchor selection used `after > before > parent`, but the insertion branch re-tested `before`, so passing both took the anchor from `after` and the insertion point from `before` — putting the section *before* the `after` target.

Rather than gate the branch on `anchor_path`, I derived the placement once alongside the anchor so the two cannot diverge again:

```python
if after:
    anchor_path, placement = after, "after"
elif before:
    anchor_path, placement = before, "before"
elif parent:
    anchor_path, placement = parent, "under"
else:
    anchor_path, placement = None, "end"
```

`placement` then drives both the default level and the insertion point; nothing downstream re-tests the raw arguments.

Added `test_after_wins_when_several_anchors_are_passed`, and verified it has teeth by reintroducing the `elif before:` form — it fails, then passes once restored.

**2. Stale docs signature.** Correct, the table row still showed the old parameter list. Updated to include `content?` and to note that `level` is inferred from the anchor when omitted, with a link to the new Section paths section.

Issue #362, opened 2026-04-24
Scheduled skills (`dream`, `mastodon-ingest`) end runs with bare status tokens instead of narrative summaries


lmorchard:
Surfaced during smoke-testing of #283 / #356.

## Observation

When I inspected scheduled-task conversation archives to feed the newsletter composer, I found that several skills end their scheduled runs with a bare status token as the last assistant message — no narrative content. The relevant pattern (verified on real archives from `workspace/conversations/schedule-*.jsonl`):

| Skill | Last assistant text |
| --- | --- |
| `linkding-ingest` | `"I have completed the scheduled task 'linkding-ingest'. Here is a summary of the processed bookmarks and their impact..."` ✓ |
| `dream` | `"HEARTBEAT_OK"` |
| `mastodon-ingest` | `"HEARTBEAT_OK"` |

`linkding-ingest` is the model: it produces a human-readable recap before the turn ends. The other two exit with just the status token.

## Impact

The newsletter (#283 / #356) reads these archives and extracts the last substantive assistant message as `final_message` per scheduled run. With nothing but `HEARTBEAT_OK`, there's no narrative substance to compose from.

#356 worked around this by adding a `_is_status_token()` filter in the newsletter parser that skips `HEARTBEAT_OK`-class tokens and falls back to the previous assistant text. That recovers something — usually the opener ("I'll begin by...") — but opener statements aren't great narrative material either.

## Proposal

Update the SKILL.md body of `dream` and `mastodon-ingest` (and any other scheduled skill with this shape) to explicitly instruct a 1-2 sentence narrative summary BEFORE the status token. Model it on what `linkding-ingest` already does.

This benefits:

- **Newsletter composition** — real material to group / prune / narrate from.
- **Human review of individual scheduled-run conversations** — the last thing a reader sees is a meaningful summary, not a framework artifact.
- **Any future retrospective / digest tool** that wants to mine these archives.

The status token itself (`HEARTBEAT_OK`) should stay — the heartbeat / status subsystem reads the token specifically, and we don't want to change that contract. The narrative goes *before* the token as a separate assistant message.

## Related

- #283 / #356 — the newsletter that motivated this observation.
- `src/decafclaw/polling.py` — shared polling preamble. Possibly the place to thread a "summarize before you finish" hint into all scheduled-skill prompts uniformly, rather than editing each SKILL.md one by one.

Issue #710, opened 2026-07-27
is_heartbeat_ok can be fooled by a mid-turn preamble on an abnormally-terminated turn


lmorchard:
Split out of #707, where it was measured but deliberately not fixed.

## The problem

`is_heartbeat_ok()` (`heartbeat.py:115-124`) returns True if the case-insensitive
substring `heartbeat_ok` appears **anywhere in the first 300 characters** of the turn's
delivered text. `polling.py:63` explicitly instructs the agent: *"If there is nothing to
report, respond with HEARTBEAT_OK."* So a mid-turn preamble mentioning the sentinel is a
plausible utterance, not a contrived one.

Heartbeat and scheduled turns have no live transport subscriber, so
`_finalize_with_note` delivers the accumulated iteration preambles alongside the
termination note (that is deliberate — the consumer only ever sees `ToolResult.text`).
If such a turn thrashes and terminates abnormally, a sentinel inside a preamble can land
in the first 300 chars and silently suppress the alert that should have surfaced.
Consumers: `heartbeat.py:182`, `schedules.py:516`.

## Current state after #707

#707 put the note **first** in the unwatched-turn join as a partial mitigation. Measured
with a realistic sentinel-bearing preamble:

| note | length | `is_heartbeat_ok` | outcome |
|---|---|---|---|
| `_finalize_loop_break` | ~336 chars | `False` | alert surfaces correctly |
| `_finalize_max_iterations` | ~65 chars | `True` | **alert still suppressed** |

So the loop-breaker path is protected (the note alone exceeds the 300-char window) and the
`max_tool_iterations` path is not — its note is far too short to push a sentinel out of
range. This is pre-existing behavior, not a #707 regression: on `main`,
`delivered = accumulated + note` unconditionally for every turn kind, so the hazard
predates that branch entirely.

## Suggested fix

Fix it at the consumer rather than by padding notes or reordering text — the real predicate
is "did this turn end normally," not "how long is the prefix." An abnormally-terminated turn
should never count as OK regardless of what its text happens to contain. Options:

1. Have `is_heartbeat_ok()` return False when the text carries an abnormal-termination
   marker (`[loop-breaker] Stopped`, `reached max tool iterations`). Cheap, but
   string-matching a different string to fix a string-matching bug.
2. Better: plumb the termination reason out of the turn rather than re-deriving it from
   prose. `_finalize_with_note` already knows the turn ended abnormally; carrying that as
[...]

PR #715, opened 2026-07-27
Pre-agent scripts for scheduled tasks, and one sentinel rule

### LINKED ISSUE #450: Pre-agent Python scripts + [SILENT] suppression for scheduled tasks
## Context

Scheduled tasks today put the prompt directly to the agent. Hermes-agent supports a pre-agent script (`script: path/to/run.py`) whose stdout becomes context the agent sees. Mechanical work (fetch RSS, query SQL, diff against last run) lives in plain Python; reasoning lives in the prompt.

Paired pattern: `[SILENT]` in the agent's response suppresses notification delivery — useful when "nothing changed, nothing to say."

## Opportunity

The newsletter skill is the obvious win. Today it has to do "fetch latest items" via tools, which is wasteful (LLM round-trips for mechanical work). A Python pre-script could pre-compute the diff and hand the agent just the new items + ask it to summarize.

Other candidates: any scheduled "monitor X, alert if changed" routine.

## Proposed approach

- Schedule frontmatter gets `pre_script: scripts/newsletter_fetch.py` (path relative to `data/{agent_id}/` or `workspace/`).
- Schedule runner executes the script in a subprocess (timeout from config), captures stdout, inserts as a `<pre_script_output>` block in the assembled prompt.
- After the agent turn and before sending the result to notification channels, check for `[SILENT]` token — if present, skip all channel delivery (still archive the turn).
- Script gets a small env (`DECAFCLAW_AGENT_ID`, `DECAFCLAW_ROUTINE_NAME`, `DECAFCLAW_WORKSPACE`); no access to decafclaw internals.

## Validation

- New test: schedule with `pre_script`, fixture script echoes `{"items": [1,2,3]}`, agent prompt contains the output.
- New test: agent response containing `[SILENT]` produces no notification, archive still written.
- Migrate newsletter skill in a follow-up.

## Open questions

- Script in Python only, or generic shell? Probably restrict to Python via `python -m` for safety.
- Should script timeout count against the routine's iteration budget? No — it's pre-agent.

## Reference

`hermes-agent/cron/scheduler.py` (script pre-injection, `[SILENT]` handling).

lmorchard:
Closes #450.

Scheduled tasks could only get data by asking the agent to fetch it — one LLM round-trip per mechanical step — and had no way to stay quiet when a cycle produced nothing worth saying.

Spec, plan, research and notes: [`docs/dev-sessions/2026-07-27-1151-450-pre-agent-scripts/`](docs/dev-sessions/2026-07-27-1151-450-pre-agent-scripts/).

## Pre-agent scripts

A schedule may declare `pre_script: scripts/fetch.py`. The runner executes it with the project interpreter before the turn and injects stdout between the preamble and the task body, so the agent reads its instructions, then the data, then what to do with it:

```
You are running a scheduled task: "feed-watch".
... preamble ...

<pre_script_output>
{
  "routine": "feed-watch",
  "new_items": [ {...}, {...} ]
}
</pre_script_output>

Summarize any new items above in one line each.
If there are no new items, reply with [SILENT] and nothing else.
```

(That block is from a real end-to-end dry run — real script, real subprocess, real assembly, mocked LLM only.)

- **Python only, never a shell.** No quoting or word-splitting surface, and schedules already have `allowed-tools: shell(...)` for shell work. Named `pre_script` rather than `pre_python` so a future `pre_command` can sit beside it without a rename.
- **Sandboxed path.** Resolved against `workspace/` then `data/{agent_id}/`, containment checked *after* resolution, so `../` escapes are rejected rather than followed.
- **Failures disclosed, not fatal.** Missing file, non-zero exit, or timeout injects an error line and the turn still runs, so the agent can report that the fetch failed. Aborting would produce silence — and silence is precisely what `[SILENT]` exists to make a *deliberate* choice about.
- **Truncation is disclosed.** Output over 8000 chars gets a notice naming the cap and the real length. A cut JSON array reads as a complete one; a silent cap would have the agent summarize a partial payload as the whole thing.
- **stderr is logged, never injected**, so a script's warnings stay out of the model's context.
- **`pre_script.timeout_sec`** (default 60s) is its own key, deliberately *not* charged against `agent.max_tool_iterations` — nothing has been inferred yet, so a slow fetch must not shrink the budget it is gathering data for.

## `[SILENT]` suppression

A scheduled response starting with `[SILENT]` skips `_notify_task_complete` entirely. Every channel adapter subscribes to the notification it emits, so **one check suppresses Mattermost, email and vault together** — no per-adapter work. The archive is untouched, per #362's requirement that scheduled archives keep real narrative for retrospective tools like `!newsletter`.

The failure path is ungated: a task that raises always notifies, whatever its response said. A suppressed run logs one line, because otherwise the absent notification is its only trace and an absence isn't diagnosable.

**Deliberately inert by default.** `build_task_preamble` is *not* changed — telling every scheduled task to consider suppressing itself is a policy change, and this is a mechanism. Task authors opt in from their own body, which is documented with an example.

## One sentinel rule

`is_heartbeat_ok` matched `HEARTBEAT_OK` *anywhere* in the first 300 characters, so a response that merely mentioned it counted as a quiet cycle. `is_background_wake_ok` was already start-anchored to avoid exactly that — two rules for one concept. And the loose one had forced unrelated code to compensate: `_finalize_with_note` orders the loop-breaker note ahead of accumulated preambles specifically so a preamble mentioning the sentinel can't bury the alert (`agent.py`, "Don't move this").

Both now delegate to `response_starts_with_sentinel`.

**Safe to tighten** because `build_task_preamble` — the only producer of these instructions — already asks for the marker first: *"respond with HEARTBEAT_OK"* (heartbeat) and *"begin your summary with HEARTBEAT_OK on its own line"* (scheduled, worded that way in #362). So this makes the code agree with the prompt. **No prompt change**, and none should be made.

The shared matcher keeps the word boundary the old regex carried ("so `BACKGROUND_WAKE_OKAY` doesn't match"), applying it only when a sentinel ends in a word character — `\b` after `]` asserts a *following* word char, which would make `"[SILENT]"` alone fail while `"[SILENT] x"` passed. The plan said `startswith`; that would have silently dropped this.

`test_is_heartbeat_ok_case_insensitive` asserted the mid-response behavior being removed, so it is rewritten rather than accommodated (CLAUDE.md: no deprecated code for test compatibility).

## Not done: no eval case

The harness can't reach this path — `eval/runner.py` never calls `run_schedule_task`, never sets `task_mode`, and `_KNOWN_SETUP_KEYS` has no schedule fixture; evals run interactive-style turns only. I established that **before** spending model calls on a case that structurally couldn't exercise the feature, and marked the plan's boxes `[!]` rather than force-ticking or dropping them.

The 19 new unit tests do assert the LLM-visible artifact — the block's presence, its payload, and its position before the task body. What they can't answer is whether a real model *uses* the injected data instead of re-fetching. `notes.md` records the follow-up that would close it: a schedule fixture in the eval harness, which would also unlock coverage for `[SILENT]`, `required-skills` and per-task tool restrictions.

## Verification

- `make test` — **3606 passed, 2 skipped** (24 net new test functions vs `origin/main`)
- `make check` — All checks passed, pyright 0 errors
- Rebased onto `origin/main` (`411bf31`, which includes #714) and re-verified after — see the conflict-resolution comment below
- Each phase was test-first with a confirmed pre-implementation failure (4 → pass, 1 → pass, 7 → pass)
- Branch self-review caught two things the phase tests didn't: the timeout path killed the child without reaping it, and truncation was silent. Both fixed with a test.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -231,6 +231,17 @@ Periodic wake-up settings.
 | `channel` | str | `""` | `HEARTBEAT_CHANNEL` |
 | `suppress_ok` | bool | `true` | `HEARTBEAT_SUPPRESS_OK` |
 
+### `pre_script`
+
+Pre-agent scripts for scheduled tasks (#450). A schedule declaring `pre_script:` runs it with the project interpreter before the turn and injects stdout into the prompt, so mechanical work costs no LLM round-trips. See [schedules.md#pre-agent-scripts](schedules.md#pre-agent-scripts).
+
+| Field | Type | Default | Env Var |
+|-------|------|---------|---------|
+| `enabled` | bool | `true` | `PRE_SCRIPT_ENABLED` |
+| `timeout_sec` | float | `60.0` | `PRE_SCRIPT_TIMEOUT_SEC` |
+
+`timeout_sec` is deliberately **not** charged against `agent.max_tool_iterations` — the script runs before any reasoning, so a slow fetch must not shrink the budget it is gathering data for. Setting `enabled = false` makes every task's `pre_script` a no-op without editing schedule files.
+
 ### `notifications`
 
 Inbox settings for the notification system (see [Notifications](notifications.md)).
--- a/docs/dev-sessions/2026-07-27-1151-450-pre-agent-scripts/notes.md
+++ b/docs/dev-sessions/2026-07-27-1151-450-pre-agent-scripts/notes.md
@@ -1,0 +1,91 @@
+# Notes — #450
+
+## Outcome
+
+Phases 1–3 implemented and green. Phase 4 (eval) **not done** — structurally
+infeasible, see below. `make test` **3599 passed, 2 skipped** (baseline 3581, so
++18 tests). `make check` clean.
+
+## Phase 4 is not doable as planned, and that was worth finding before spending
+
+The plan called for an eval case asserting a model uses injected `pre_script`
+output instead of re-fetching. The harness can't express it: `eval/runner.py`
+never calls `run_schedule_task`, never sets `task_mode`, and `_KNOWN_SETUP_KEYS`
+(`runner.py:491`) has no schedule fixture. Evals run interactive-style turns only.
+
+A case would need a new harness capability — a schedule fixture that drives a task
+through the scheduled path — which is larger than #450 and is *not* something the
+generic `config_overrides` mechanism reaches (CLAUDE.md explicitly warns against
+adding bespoke `setup.*` keys, but this would be a new capability, not a knob).
+
+I checked this before writing the case rather than after, so no model calls were
+spent on something that couldn't have exercised the feature.
+
+**What covers it instead:** the 8 Phase 3 unit tests assert the LLM-visible
+artifact directly — the block's presence, its payload, and its position before the
+task body. What they can't answer is the behavioral question (does a real model
+actually use the data rather than re-fetch). That gap is real and unclosed.
+
+**Candidate follow-up issue:** teach the eval harness to run a scheduled task
+(schedule fixture + `run_schedule_task` path). Would unlock eval coverage for
+`pre_script`, `[SILENT]`, `required-skills`, and per-task tool restrictions — none
+of which are eval-reachable today.
+
+## Two plan assertions turned out wrong
+
+Both marked `[!]` in `plan.md` rather than force-ticked:
+
+1. **`startswith` was the wrong implementation.** The plan specified it, but the
+   existing `_BACKGROUND_WAKE_OK_RE` carried `\b` "so BACKGROUND_WAKE_OKAY doesn't
+   match" — `startswith` silently loses that. The shared matcher compiles a cached
+   regex and applies the boundary only when the sentinel ends in a word character;
+   it can't be applied to `[SILENT]`, since `\b` after `]` would demand a
+   following word char and `"[SILENT]"` alone would fail. Caught by writing a
+   word-boundary test that the plan didn't have.
+2. **The durations assertion was written against a design that changed.**
+   `test_pre_script_timeout_is_disclosed` *is* the slowest test in its file — but
+   at 0.11s, because `timeout_sec` ended up `float` and the test uses `0.05`. The
+   plan assumed a 1s timeout. Intent (no fixed multi-second waits) holds.
+
+## Interaction worth knowing
+
+`[SILENT]` and `HEARTBEAT_OK` are independent sentinels. A quiet cycle that
+replies `[SILENT]` yields `is_ok=False` in `run_schedule_task`'s return dict,
+because `is_ok` is specifically the `HEARTBEAT_OK` measure. Harmless today — the
+notification that would have used `is_ok` for its title and priority is the exact
+thing being suppressed. It would matter if someone later made suppression
+conditional; if so, decide then whether `[SILENT]` should imply `ok`.
+
+## End-to-end dry run
+
+Real script, real subprocess, real assembly (mocked LLM only):
+
+```
+You are running a scheduled task: "feed-watch".
+... preamble ...
+
+<pre_script_output>
+{
+  "routine": "feed-watch",
+  "new_items": [ {...}, {...} ]
+}
+</pre_script_output>
+
+Summarize any new items above in one line each.
+If there are no new items, reply with [SILENT] and nothing else.
+```
+
+`notifications sent: 0` (suppressed), response preserved in the return dict,
+`DECAFCLAW_ROUTINE_NAME` visibly reached the script.
+
+## Scope held
+
+Newsletter migration, a generic `pre_command`, post-turn hooks, `build_task_preamble`
+wording, and `agent.py`'s note ordering were all named as out of scope in the spec
+and none were touched. The `agent.py` change is comment-only.
+
+## Size
+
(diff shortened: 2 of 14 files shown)


lmorchard:
Addressed the Copilot review — **4 of 5 fixed, 1 adopted for a different reason than given**. Each fix landed with a test that failed first; all four findings reproduced.

## Fixed

**1. `env={**os.environ, ...}` leaked the whole parent environment.** The most serious of the five, and a straight deviation from what #450 specified ("a small env"). Provider credentials live in this process, and a `pre_script`'s stdout goes directly into model context — so `print(os.environ)` in an otherwise innocent script would exfiltrate them. Now an explicit allowlist (`PATH`, `HOME`, locale, proxy, TLS bundle vars) plus the three documented `DECAFCLAW_*`. Test sets a fake `VERTEX_SECRET_TOKEN` and asserts it never reaches the prompt.

**2. A payload containing `</pre_script_output>` broke the block open.** Real, and worse than "confuses parsing": everything after the injected closing tag would read as *prompt text* rather than data. Neutralized. Test emits the closing tag followed by `Ignore previous instructions.` and asserts the prompt contains exactly one closing tag and that the injected text stays inside the block.

**3. `write_overlay` documented `pre_script` as a patch key but never applied it.** Correct — I updated the docstring and missed the `replace(...)` call, so the docs lied and UI patches couldn't set the path. Best catch of the five, because my own round-trip test *passed* while the bug existed: it only asserted an unrelated patch **preserves** the field, which works via `base` even when the key is ignored. Added `test_write_overlay_can_set_pre_script`, which failed before the fix.

**4. Empty sentinel matched everything.** `response_starts_with_sentinel(x, "")` compiles to `^\s*`, so a config-driven empty value would have suppressed every response. Guarded, with a test covering `""` and `"   "`.

## Adopted, but the stated rationale is wrong

**5. `except TimeoutError` → `except asyncio.TimeoutError`.** Changed, for codebase consistency — `mcp_client.py`, `polling.py` and `runner.py` all use the qualified form, and mine was the only bare one.

But the reason given ("avoids accidentally catching unrelated `TimeoutError`s") doesn't hold: since Python 3.11 `asyncio.TimeoutError` **is** the builtin. Verified on this interpreter:

```
$ python -c "import asyncio; print(asyncio.TimeoutError is TimeoutError)"
True     # 3.13.14
```

So the change is cosmetic. Worth making, not worth believing it fixed a bug.

## Verification

`make test` **3604 passed, 2 skipped** (was 3600 pre-review; +4 tests). `make check` clean. Rebased and re-verified; squashed and force-pushed with `--force-with-lease`.

lmorchard:
## Rebased onto `411bf31` — conflict resolution

`origin/main` gained #714 ("Never report an abnormally-terminated turn as heartbeat-OK") while this branch was open. Both commits edit `is_heartbeat_ok`, so `src/decafclaw/heartbeat.py` conflicted; `docs/schedules.md` and `tests/test_heartbeat.py` auto-merged.

The two changes are complementary, not competing, so the resolution keeps both:

- **#714's `_ABNORMAL_TERMINATION_MARKERS` override** — whole-response, case-sensitive, checked first.
- **#450's start-anchored `response_starts_with_sentinel`** — now what the marker check falls through to.

```python
if response and any(m in response for m in _ABNORMAL_TERMINATION_MARKERS):
    return False
return response_starts_with_sentinel(response, "HEARTBEAT_OK")
```

Start-anchoring does **not** subsume the override: "did this turn end normally" is a different predicate from where the sentinel sits. Should a sentinel-bearing preamble ever reach the front of the response, the start-anchored match fires and buries the alert. That's why the override stays, and why it stays whole-response-scoped.

### #714's regression guard had lost its teeth — rewritten

Worth flagging, because the rebase produced an all-green suite that was quietly weaker than either parent.

`test_is_heartbeat_ok_false_on_abnormal_termination` used note-first fixtures (`"\n\n[Agent reached max tool iterations…]"` then the sentinel). Under the tightened matcher those responses no longer start with `HEARTBEAT_OK`, so they return `False` via start-anchoring alone. I verified this by deleting the override and re-running: **40/40 still passed.** The guard pinning #714 no longer discriminated the code it was written to pin.

Fixed by inverting the fixtures — sentinel first, marker after — which is the only arrangement the whole-response marker check can catch, plus a premise guard so the test fails loudly if a future change makes it incidental again:

```python
assert response_starts_with_sentinel(text, "HEARTBEAT_OK") is True, (
    f"{label}: fixture no longer exercises the override"
)
assert is_heartbeat_ok(text) is False
```

Added `test_is_heartbeat_ok_abnormal_marker_matched_beyond_scan_window`, pinning the whole-response scope that #714 recorded as covered by no check.

Re-probed after: with the override deleted, **both** guards now fail. Restored, all pass.

### Docs

`docs/heartbeat.md` was added by #714 and so never conflicted, but its `HEARTBEAT_OK` section still described the old "anywhere within the first 300 characters" rule. Updated to the start-anchored rule, and the abnormal-termination paragraph now explains why start-anchoring narrows the window without closing it. `agent.py`'s ordering comment now cites both #710 and #450 as the reasons it's belt-and-braces.

### Verification

- `make test` — 3606 passed, 2 skipped
- `make check` — ruff clean, pyright 0 errors, tsc clean, message-types drift check clean
- No behavior change to this PR's own features; the resolution is confined to the sentinel path.

PR #720, opened 2026-07-27
docs: preserve the parked #471 session docs

lmorchard:
Rescues four session docs that existed **only** as untracked files in a local worktree — not on `main`, and not in PR #518 either, since they were never committed. Cleaning up that worktree would have destroyed them.

Found while clearing out stale worktrees after #715 merged.

## Why keep them

PR #518 was [closed without merge](https://github.com/lmorchard/decafclaw/pull/518) over a framing-vs-reality gap: the "sandbox" was a subprocess runner with stdlib unrestricted, so a script could `import subprocess` and bypass the shell-confirmation gate, `import urllib.request` and bypass the `http_request` gate, or `import os` and bypass workspace-write guards.

**Issue #471 is still open.** The spec and that architectural finding are the starting point for whoever picks it up, and re-deriving them costs a full brainstorm-and-plan cycle.

Three findings in `notes.md` are independent of the parked design and still hold:

- macOS `setrlimit(RLIMIT_AS, ...)` raises `ValueError: current limit exceeds maximum limit` — the kernel refuses to lower `RLIM_INFINITY`. The plan had assumed it was silently best-effort.
- `tools/delegate.py` publishes `tool_status` with a **positional dict** where consumers read `**kwargs` off the top-level event, so the payload is silently dropped.
- `copy.copy(parent_ctx)` shares mutable subobjects, so concurrent tool calls in one parent turn can race on `ctx.tools`.

The `delegate.py` one looks like a live bug on `main` and is worth its own issue — I did not touch it here.

## Notes

- Added a **status header** to `notes.md` so the 680-line plan doesn't read as shipped behavior. No `code_execution` skill exists on `main`, and nothing in these docs describes current code.
- Files are otherwise byte-identical to the worktree copies (verified with `diff -rq` before committing).
- Docs only — no code, no tests. Nothing for CI to exercise beyond the lint pass.

### DIFF
--- a/docs/dev-sessions/2026-05-15-1244-programmatic-tool-calling-471/notes.md
+++ b/docs/dev-sessions/2026-05-15-1244-programmatic-tool-calling-471/notes.md
@@ -1,0 +1,27 @@
+# Session notes
+
+## Status: parked, not shipped
+
+PR [#518](https://github.com/lmorchard/decafclaw/pull/518) was **closed without merge** on 2026-05-17. The second Copilot review surfaced a framing-vs-reality gap: the "sandbox" was a subprocess runner with stdlib unrestricted, so a script could `import subprocess` and bypass the shell-confirmation gate, `import urllib.request` and bypass the `http_request` gate, or `import os` and bypass workspace-write guards. Issue [#471](https://github.com/lmorchard/decafclaw/issues/471) remains **open**.
+
+Read the plan below as a record of an attempt, not a description of shipped behavior — no `code_execution` skill exists on `main`. The technical findings further down (macOS `RLIMIT_AS`, the `delegate.py` wire-format bug, the `copy.copy` fork race) are independent of the parked design and still hold.
+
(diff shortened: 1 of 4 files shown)


lmorchard:
Amended after checking the branch history: these docs are **incomplete**, not just parked.

They were last written 2026-05-15 14:50, but the branch continued for seven commits through 2026-05-16 12:44 — allowlisting `workspace_glob`/`workspace_search`, auto-injecting the dc-proxy import, and several `ToolResult.data` corrections. No retro was written before #518 was parked, so `notes.md` stops mid-stream at Phase 4.

The status header now says so and points at `origin/programmatic-tool-calling-471` (`00c51ae`) for the actual final state. **That remote branch should stay** — it holds seven commits that exist nowhere else, and it's the only complete record of the attempt while #471 is open.

Also worth knowing: the local worktree I rescued these from was six commits *behind* that remote, which is why the docs stop where they do.

